PR #2, opened 2024-03-09
WIP: Talents and Glyphs

ToxicKevinFerm:
Fixed talent and glyph generators.
Had to comment out old wotlk sims so they're pretty gutted.

* Talent and Glyph generators fixed for Cata, both for player and hunter pets
* Implemented Prime glyphs
* Commented our wotlk sims

**Note: Doesn't build Typescript in this configuration because of old sim TS presets. Just here for WIP**


rosenrusinov:
I think we want to merge #3 first then re-run the updated scrapers on the new spec setups

ToxicKevinFerm:
> I think we want to merge #3 first then re-run the updated scrapers on the new spec setups

Definitely waiting on #3 - I anticipate a few changes at least although the scrapers are mostly not spec dependant

ToxicKevinFerm:
> I think we want to merge #3 first then re-run the updated scrapers on the new spec setups

Though maybe we want to change the class proto to instead of carrying every talent in the DruidTalents message, maybe we want to split it into DruidRestorationTalents etc? @rosenrusinov

ToxicKevinFerm:
Also to make the proto generation easier, could we split it into folders and file per spec maybe?
Say
```
/proto
  /classes
    /druid
      druid.generated.proto
      restoration.proto
      feraldps.proto
      feraltank.proto
      balance.proto
```
So generated stuff is in generated.proto. Generally not a fan of mixing generated and non generated in the same file.

PR #15, opened 2024-03-16
Remove unused stats + Fix item generation

kayla-glick:


rosenrusinov:
I think you should update the indices to rekove the gap made by removing energy and rage. Its the only chance we will have to do this. Also need to uodate the tools stat array to match to make sure items stats work correct i believe

kayla-glick:
Oh true good callout @rosenrusinov

PR #17, opened 2024-03-16
Hunter + Focus

ToxicKevinFerm:
Impelements FocusBar as a resource, with a base in Energy,

Ready to be merged.

Adds

1. Focus as Hunter resource
2. BM Mastery
3. MM Mastery
4. SV Mastery
5. All* talents (might have missed one, and disregarding the no dps ones)
6. All* Glyphs (dps related, could add more for sim purpose maybe)
7. All new DPS abilities
8. T11 Set bonus
9. Everything uses Focus

I want to start working on stuff around it, like adding potions etc. and it'd be easier if we got this merged so other people can look at it too.
Not 100% by any means, but it's a good way there.


sim/core/character.go
```diff
@@ -466,6 +466,13 @@ func (character *Character) Finalize() {
 		})
 	}
 
+	if character.Class == proto.Class_ClassHunter {
+		character.Env.RegisterPostFinalizeEffect(func() {
+			character.focusBar.setupFocusThresholds()
+		})
+	}
```
rosenrusinov, line 473: Does this have to be here and not in the class file? I see the rogue one was also here but we can probably move it out as well? I'm not a fan of class specific code in the core

sim/core/focus.go
```diff
@@ -2,41 +2,120 @@ package core
...
 	nextFocusTick time.Duration
+	isPlayer bool
+
+	// Multiplies focus regen from ticks.
+	FocusTickMultiplier float64
 
-	regenMetrics  *ResourceMetrics
-	refundMetrics *ResourceMetrics
+	regenMetrics        *ResourceMetrics
+	FocusRefundMetrics *ResourceMetrics
```
rosenrusinov, line 38: do you need this to be public?

sim/core/focus.go
```diff
@@ -2,41 +2,120 @@ package core
...
-		onFocusGain:   onFocusGain,
-		regenMetrics:  unit.NewEnergyMetrics(ActionID{OtherID: proto.OtherAction_OtherActionFocusRegen}),
-		refundMetrics: unit.NewEnergyMetrics(ActionID{OtherID: proto.OtherAction_OtherActionRefund}),
+		unit:                 unit,
+		maxFocus:            max(100, maxFocus),
+		FocusTickMultiplier: 1,
+		isPlayer: isPlayer,
+		baseFocusPerSecond: baseFocusPerSecond,
+		regenMetrics:         unit.NewFocusMetrics(ActionID{OtherID: proto.OtherAction_OtherActionFocusRegen}),
+		FocusRefundMetrics:  unit.NewFocusMetrics(ActionID{OtherID: proto.OtherAction_OtherActionRefund}),
```
rosenrusinov, line 51: it looks like your auto formating is not running properly I see weird indents here

sim/core/focus.go
```diff
@@ -73,85 +184,91 @@ func (fb *focusBar) SpendFocus(sim *Simulation, amount float64, metrics *Resourc
 	}
 
 	newFocus := fb.currentFocus - amount
-	metrics.AddEvent(-amount, -amount)
+	//metrics.AddEvent(-amount, -amount)
 
 	if sim.Log != nil {
-		fb.unit.Log(sim, "Spent %0.3f focus from %s (%0.3f --> %0.3f).", amount, metrics.ActionID, fb.currentFocus, newFocus)
+		//fb.unit.Log(sim, "Spent %0.3f focus from %s (%0.3f --> %0.3f).", amount, metrics.ActionID, fb.currentFocus, newFocus)
```
rosenrusinov, line 190: i think commenting this will remove focus from the timeline as the timeline is parsed from logs

sim/core/focus.go
```diff
@@ -73,85 +184,91 @@ func (fb *focusBar) SpendFocus(sim *Simulation, amount float64, metrics *Resourc
 	}
 
 	newFocus := fb.currentFocus - amount
-	metrics.AddEvent(-amount, -amount)
+	//metrics.AddEvent(-amount, -amount)
```
rosenrusinov, line 187: Why is this commented out?

sim/hunter/arcane_shot.go
```diff
@@ -0,0 +1,46 @@
...
+		},
+
+		BonusCritRating: 0,
+		DamageMultiplierAdditive: 1,
+		DamageMultiplier: dmgMultiplier,
+		CritMultiplier: hunter.CritMultiplier(true, true, false),
+		ThreatMultiplier: 1,
+
+		ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
+			wepDmg := hunter.AutoAttacks.Ranged().CalculateWeaponDamage(sim, spell.RangedAttackPower(target))
```
rosenrusinov, line 39: Not normalized? Just asking to make sure

sim/core/focus.go
ToxicKevinFerm: Youre right, it wanted to use prettier for some reason because of a config file in root.. Fixing

sim/hunter/aspects.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+// Todo: Implement Aspect of the Fox?
+
+func (hunter *Hunter) applySharedAspectConfig(isHawk bool, aura *core.Aura) {
+	if isHawk {
+		aura.OnReset = func(aura *core.Aura, sim *core.Simulation) {
+			aura.Activate(sim)
+		}
+	}
```
rosenrusinov, line 38: I would remove this part as it was for the old non APL rotations. Apl should now specify the Aspect cast in the Pre-pull actions

sim/hunter/cata_items.go
```diff
@@ -0,0 +1,47 @@
...
+			})
+		},
+		4: func(agent core.Agent) {
+			hunter := agent.(HunterAgent).GetHunter()
+			hunter.RegisterAura(core.Aura{
+				Label:    "T11 4pc",
+				Duration: core.NeverExpires,
+				OnReset: func(aura *core.Aura, sim *core.Simulation) {
+					aura.Activate(sim)
+				},
```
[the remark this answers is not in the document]
rosenrusinov: Looking at this im not sure i like this implementation. Cant you check for the set in the registration of those spells and just assing the correct cast time if the set is equipped

sim/hunter/cata_items.go
```diff
@@ -0,0 +1,47 @@
...
+				OnGain: func(aura *core.Aura, sim *core.Simulation) {
+					hunter.SerpentSting.BonusCritRating += 0.05
+				},
+				OnExpire: func(aura *core.Aura, sim *core.Simulation) {
+					hunter.SerpentSting.BonusCritRating -= 0.05
+				},
```
rosenrusinov, line 25: same comments as the 4pc bonus

rosenrusinov: also when working with `BonusCritRating` its prefered to do `N * core.CritRatingPerCritChange`

sim/hunter/hunter.go
```diff
@@ -79,53 +92,50 @@ func (hunter *Hunter) GetHunter() *Hunter {
...
+	}
+	if hunter.Talents.FerociousInspiration && hunter.Pet != nil {
+		raidBuffs.FerociousInspiration = true
+	}
 }
 func (hunter *Hunter) AddPartyBuffs(_ *proto.PartyBuffs) {
 }
+func (hunter *Hunter) CalculateMasteryPoints() float64 {
+	return hunter.GetStat(stats.Mastery) / core.MasteryRatingPerMasteryPoint
+}
```
rosenrusinov, line 106: why is there mastery related stuff in the core hunter?

sim/hunter/cata_items.go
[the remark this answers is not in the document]
ToxicKevinFerm: > Looking at this im not sure i like this implementation. Cant you check for the set in the registration of those spells and just assing the correct cast time if the set is equipped

I wasn't sure that was a thing, but definitely

sim/hunter/hunter.go
```diff
@@ -139,53 +149,34 @@ func NewHunter(character *core.Character, options *proto.Player, hunterOptions *
...
+		OffHand:         hunter.WeaponFromOffHand(0),
+		Ranged:          rangedWeapon,
+		//ReplaceMHSwing:  hunter.TryRaptorStrike, //Todo: Might be weaving
+		AutoSwingRanged: true,
+	})
+
+	hunter.AutoAttacks.RangedConfig().ApplyEffects = func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
+		baseDamage := hunter.RangedWeaponDamage(sim, spell.RangedAttackPower(target)) +
+			spell.BonusWeaponDamage()
+		spell.CalcAndDealDamage(sim, target, baseDamage, spell.OutcomeRangedHitAndCrit)
```
rosenrusinov, line 176: do you need to do this if you're not changing it from the default implementation? Wotlk/Sod do this to add the Ammo bonus damage which seems to be gone now

sim/hunter/cata_items.go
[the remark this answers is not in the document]
rosenrusinov: Yeah we have a method that can be called during registration (make sure to cache the value if you want to use is in ApplyEffects as its expensive) `class.HasSetBonus(ItemSetWhatever, 4)`

sim/hunter/hunter.go
ToxicKevinFerm: So I can use it more than once in the upper parts. I can remove it np though

sim/core/focus.go
rosenrusinov: which makes me think you probably want to look into the TS files to see where we parse resource metrics if you have to add something new for focus (even tho focus is not a new resource so should be there)

sim/hunter/kill_command.go
```diff
@@ -0,0 +1,52 @@
...
+				GCD: time.Second,
+			},
+			CD: core.Cooldown{
+				Timer:    hunter.NewTimer(),
+				Duration: time.Second * 6,
+			},
+		},
+		ExtraCastCondition: func(sim *core.Simulation, target *core.Unit) bool {
+			return true
+		},
```
rosenrusinov, line 38: not needed if it just returns true

sim/hunter/kill_command.go
```diff
@@ -0,0 +1,52 @@
...
+		DamageMultiplierAdditive: 1 + (float64(hunter.Talents.ImprovedKillCommand) * 0.05),
+		CritMultiplier:   hunter.CritMultiplier(false, false, false),
+		ThreatMultiplier: 1,
+		ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
+			if hunter.Talents.KillingStreak > 0 {
+				if !hunter.KillingStreakCounterAura.IsActive() {
+					hunter.KillingStreakCounterAura.Activate(sim)
+				}
```
rosenrusinov, line 46: Why are you checking if the aura is not active? Should this not refresh the duration of the aura? Activate already checks if the aura is active and refreshes instead

sim/hunter/kill_shot.go
```diff
@@ -0,0 +1,79 @@
...
+
+import (
+	"time"
+
+	"github.com/wowsims/cata/sim/core"
+	"github.com/wowsims/cata/sim/core/proto"
+)
+
+func (hunter *Hunter) registerKillShotSpell() {
+	if hunter.HasPrimeGlyph(proto.HunterPrimeGlyph_GlyphOfKillShot) {
```
rosenrusinov, line 11: when doing registers based on a condition we usually prefer the Early Exit Pattern:
```
func (hunter *Hunter) registerKillShotSpell() {
	if hunter.HasPrimeGlyph(proto.HunterPrimeGlyph_GlyphOfKillShot) {
        	return;
	}
        spellImpl...
```

rosenrusinov: ahh i just saw this is an extra condition on the spell and not the spell itself you can ignore

sim/hunter/kill_shot.go
```diff
@@ -0,0 +1,79 @@
...
+			Duration: time.Second * 6,
+		}
+		ksReset := hunter.RegisterAura(core.Aura{
+			Label: "Kill Shot Glyph",
+			Icd: &icd,
+			Duration: time.Second,
+			OnGain: func(aura *core.Aura, sim *core.Simulation) {
+				hunter.KillShot.CD.Reset()
+			},
+		})
```
rosenrusinov, line 23: What is the reason this is implemented as a 1 second aura instead of ust calling `hunter.KillShot.CD.Reset()` instead of applying an aura that does it

sim/hunter/kill_shot.go
```diff
@@ -0,0 +1,79 @@
...
+			Label: "Kill Shot Glyph Activator",
+			Duration: core.NeverExpires,
+			OnReset: func(aura *core.Aura, sim *core.Simulation) {
+				aura.Activate(sim)
+			},
+			OnSpellHitDealt: func(aura *core.Aura, sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
+				if spell == hunter.KillShot {
+					if ksReset.Icd.IsReady(sim) {
+						ksReset.Icd.Use(sim)
+						ksReset.Activate(sim)
```
rosenrusinov, line 34: you can just call `hunter.KillShot.CD.Reset()` instead of applying this aura. The only reason for an aura i would see is to show this in the metrics but maybe we can do this smarter as well as it feels a bit overkill to do it with an aura

sim/hunter/kill_shot.go
```diff
@@ -0,0 +1,79 @@
...
+		CritMultiplier: 1,//  hunter.critMultiplier(true, true, false),
+		ThreatMultiplier: 1,
+		// https://web.archive.org/web/20120207222124/http://elitistjerks.com/f74/t110306-hunter_faq_cataclysm_edition_read_before_asking_questions/
+		ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
+			// (100% weapon dmg + 45% RAP + 543) * 150%
+			normalizedWeaponDamage := hunter.AutoAttacks.Ranged().CalculateNormalizedWeaponDamage(sim, spell.RangedAttackPower(target))
+			rapBonusDamage := spell.RangedAttackPower(target) * 0.45
+			flatBonus := 543.0
+
+			baseDamage := (normalizedWeaponDamage + rapBonusDamage + flatBonus) * 1.5
```
rosenrusinov, line 75: the ` * 1.5` should be moved out of the dynamic multipliers into the static `DamageMultiplier` of the spell config

sim/hunter/marksmanship/aimed_shot.go
```diff
@@ -0,0 +1,58 @@
...
+func (hunter *MarksmanshipHunter) registerAimedShotSpell(timer *core.Timer) {
+	if hunter.HasPrimeGlyph(proto.HunterPrimeGlyph_GlyphOfAimedShot) {
+		focusMetrics := hunter.NewFocusMetrics(core.ActionID{SpellID: 42897})
+		hunter.RegisterAura(core.Aura{
+		OnSpellHitDealt: func(aura *core.Aura, sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
+			if spell == hunter.AimedShot && result.DidCrit() {
+				hunter.AddFocus(sim, 5, focusMetrics)
+			}
+		},
+		})
```
rosenrusinov, line 19: i might be remembering wrong but i think every aura needs atleast a `Label`

sim/hunter/marksmanship/chimera_shot.go
```diff
@@ -20,31 +20,27 @@ func (hunter *Hunter) registerChimeraShotSpell() {
...
+		DamageMultiplier: 1,
+		CritMultiplier:   hunter.CritMultiplier(true, true, false),
 		ThreatMultiplier: 1,
 
 		ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
 			baseDamage := 0.2*spell.RangedAttackPower(target) +
 				hunter.AutoAttacks.Ranged().BaseDamage(sim) +
-				hunter.AmmoDamageBonus +
 				spell.BonusWeaponDamage()
 			baseDamage *= 1.25
```
rosenrusinov, line 45: this `1.25` multiplier can be moved to the static spell DamageMultiplier instead of here

sim/hunter/marksmanship/chimera_shot.go
```diff
@@ -62,20 +58,18 @@ func (hunter *Hunter) registerChimeraShotSpell() {
 	})
 }
 
-func (hunter *Hunter) chimeraShotSerpentStingSpell() *core.Spell {
+func (hunter *MarksmanshipHunter) chimeraShotSerpentStingSpell() *core.Spell {
```
rosenrusinov, line 61: I'm not sure how this spell works in cata so i might be wrong but isnt this supposed to do % of the total damage of Serpent Sting? You can check the SOD implementation for how we do it there

sim/hunter/marksmanship/marksmanship.go
```diff
@@ -22,20 +22,66 @@ func RegisterMarksmanshipHunter() {
...
 	}
 	mmHunter.MarksmanshipOptions = mmOptions
-
 	return mmHunter
 }
+func (mmHunter *MarksmanshipHunter) Initialize() {
+	mmHunter.Hunter.Initialize()
+
+	mmHunter.aimedShotTimer = mmHunter.NewTimer()
+	mmHunter.registerAimedShotSpell(mmHunter.aimedShotTimer)
```
rosenrusinov, line 76: the only reason for passing this timemr would be if you want to pass it to another spell and make them share CD. If its used in a single spell its better to not pass anything and create a new timer in the spell registration

rosenrusinov: If you want to reset the timer of the AimedShot then you can just do `hunter.AimedShot.CD.Reset()` so you dont need to hold a reference to this timer in the MM struct

sim/hunter/marksmanship/marksmanship.go
```diff
@@ -22,20 +22,66 @@ func RegisterMarksmanshipHunter() {
...
+	mmHunter.registerChimeraShotSpell()
+	mmHunter.applyMastery()
+}
 
 type MarksmanshipHunter struct {
 	*hunter.Hunter
+
+	aimedShotTimer *core.Timer
```
rosenrusinov, line 84: above comment applies to this reference. Is it really needed?

sim/hunter/arcane_shot.go
ToxicKevinFerm: https://www.wowhead.com/cata/spell=3044/arcane-shot Little bit unclear here. The modifier seems to be not normalized, but effect 1 is. Idk which to use here, gotta revisit it

sim/hunter/rapid_fire.go
```diff
@@ -24,13 +24,12 @@ func (hunter *Hunter) registerRapidFireCD() {
 		OnGain: func(aura *core.Aura, sim *core.Simulation) {
 			aura.Unit.MultiplyRangedSpeed(sim, hasteMultiplier)
 
-			if manaMetrics != nil {
-				manaPerTick := 0.02 * float64(hunter.Talents.RapidRecuperation) * hunter.MaxMana()
+			if focusMetrics != nil {
 				core.StartPeriodicAction(sim, core.PeriodicActionOptions{
```
rosenrusinov, line 28: I know you havent changed this but its a bug fix i did for other periodic actions like this one.
define a PA var above the aura and in the OnGain assing the periodic action you start to this var and on the expire do `pa.Cancel(sim)` to make sure if for some reason the aura is removed before its original expiration time it properly stops the regen as well (I had to fix a couple of those on the SoD fork)

sim/hunter/marksmanship/chimera_shot.go
ToxicKevinFerm: Actually it doesnt in Cata, shouldnt be there at all, removing it.

sim/hunter/steady_shot.go
```diff
@@ -0,0 +1,59 @@
...
+				}
+			}
+			baseDamage := hunter.AutoAttacks.Ranged().CalculateWeaponDamage(sim, spell.RangedAttackPower(target)) + (280 + spell.RangedAttackPower(target) * 0.021)
+			focus := 9.0
+			if hunter.Talents.Termination != 0 && sim.IsExecutePhase25() {
+				focus = float64(hunter.Talents.Termination) * 3
+			}
+			hunter.AddFocus(sim, focus, ssMetrics)
+			result := spell.CalcDamage(sim, target, baseDamage, spell.OutcomeRangedHitAndCrit)
+			spell.DealDamage(sim, result)
```
rosenrusinov, line 56: If you're not doing anything with the result you can just do `spell.CalcAndDealDamage`.

Now that I look at this lots of your ranged spells are missing travel time. You can check how to implement that from the SOD fork (I did implement them all with travel time even if the missile speed is not entirely correct)

sim/hunter/survival/black_arrow.go
```diff
@@ -0,0 +1,74 @@
...
+				Label: "Black Arrow Dot",
+			},
+			NumberOfTicks: 10,
+			TickLength:    time.Second * 2,
+			OnSnapshot: func(sim *core.Simulation, target *core.Unit, dot *core.Dot, isRollover bool) {
+				// https://web.archive.org/web/20120207222124/http://elitistjerks.com/f74/t110306-hunter_faq_cataclysm_edition_read_before_asking_questions/
+				//  66.5% RAP + 2849 (total damage) - changed 6/28 in 4.2 (based off spell crit multiplier, modified by toxicology)
+				// https://wago.tools/db2/SpellEffect?build=4.4.0.53750&filter[SpellID]=exact%3A3674&page=1
+				baseDamage := 2849.0
+				rap := dot.Spell.Unit.GetStat(stats.RangedAttackPower)
```
rosenrusinov, line 53: You should use the dedicated function in spell for getting any AttackPower (be it melee or ranged) as it includes extra bonuses you are currently missing when you just grab the stat

sim/hunter/survival/black_arrow.go
```diff
@@ -0,0 +1,74 @@
...
+				percentageOfRAP := 0.665
+
+				// SnapshotBaseDamage calculation for the DoT, divided by 10 to spread across all ticks
+				dot.SnapshotBaseDamage = baseDamage + (percentageOfRAP * rap) / 10
+
+				dot.SnapshotAttackerMultiplier = dot.Spell.AttackerDamageMultiplier(dot.Spell.Unit.AttackTables[target.UnitIndex])
```
rosenrusinov, line 59: correct me if I'm wrong but dont dots in cata no longer snapshot attack/crit multipliers and update them on every tick?

sim/hunter/survival/explosive_shot.go
```diff
@@ -0,0 +1,74 @@
...
+		ThreatMultiplier: 1,
+
+		Dot: core.DotConfig{
+			Aura: core.Aura{
+				Label: "Explosive Shot - Dot",
+			},
+			NumberOfTicks: 2,
+			TickLength:    time.Second * 1,
+			OnSnapshot: func(sim *core.Simulation, target *core.Unit, dot *core.Dot, isRollover bool) {
+				rap := dot.Spell.Unit.GetStat(stats.RangedAttackPower) + dot.Spell.Unit.PseudoStats.MobTypeAttackPower
```
rosenrusinov, line 46: use `spell.RangedAttackPower` instead of manually trying to build the RAP

sim/hunter/survival/explosive_shot.go
```diff
@@ -0,0 +1,74 @@
...
+			if result.Landed() {
+				spell.SpellMetrics[target.UnitIndex].Hits--
+				if spell.Dot(target).IsActive() { // Todo: Implement rollover of last tick in sim.go
+					spell.Dot(target).Rollover(sim)
+					spell.Dot(target).TickOnce(sim)
+				} else {
+					dot := spell.Dot(target)
+					dot.Apply(sim)
+					dot.TickOnce(sim)
+				}
```
rosenrusinov, line 70: I would not try to emulate the new dot refresh here and just use the standard `dot.Apply` and leave the core to handle the rest when we update it

ui/core/components/individual_sim_ui/apl_values.ts
```diff
@@ -606,6 +607,13 @@ const valueKindFactories: { [f in NonNullable<APLValueKind>]: ValueKindConfig<AP
 		newValue: APLValueCurrentRage.create,
 		fields: [],
 	}),
+	currentFocus: inputBuilder({
+		label: 'Focus',
+		submenu: ['Resources'],
+		shortDescription: 'Amoiunt of currently available Focus.',
```
rosenrusinov, line 613: typo

ui/hunter/beast_mastery/sim.ts
```diff
@@ -118,7 +118,7 @@ const SPEC_CONFIG = registerSpecConfig(Spec.SpecBeastMasteryHunter, {
 	},
 
 	// IconInputs to include in the 'Player' section on the settings tab.
-	playerIconInputs: [HunterInputs.PetTypeInput(), HunterInputs.WeaponAmmo(), HunterInputs.UseHuntersMark()],
+	playerIconInputs: [HunterInputs.PetTypeInput(), HunterInputs.UseHuntersMark()],
```
rosenrusinov, line 121: Do we want to have Hunters Mark be a player input? Isnt it better if its an actual spell that can be cast with a prepull action?

sim/hunter/survival/black_arrow.go
ToxicKevinFerm: I'm not an expert on dots by any means but I think the change is that they always snapshot?
https://cynwise.wordpress.com/2011/11/18/how-warlock-dots-work-in-cataclysm-part-ii-dot-refreshing-proc-stacking-and-getting-your-warlock-black-belt/
Unless I'm misunderstanding something

sim/core/focus.go
ToxicKevinFerm: No the focus timeline works, idk why that one is needed but i can add it back np. Had some lag issues with the resource metrics before I think thats why I commented them out

ToxicKevinFerm: I dont have them now though. Thats the issue with old and big branches :D Things are not the same as when I started. I'll just add them back in

sim/core/focus.go
ToxicKevinFerm: Fixed

sim/hunter/aspects.go
ToxicKevinFerm: Removed, not sure why it was there or how to test but it applies with APL at least

sim/hunter/hunter.go
ToxicKevinFerm: Not sure how the default implementation works. but I removed it for now, the outcome seems to be the same.

sim/hunter/kill_command.go
ToxicKevinFerm: Removed

sim/hunter/kill_shot.go
ToxicKevinFerm: Had no idea Cooldowns worked without auras tbh. Then I can prolly change my counter auras in talents.go as well, take a look at the new implementation and see if I understood you correctly.

sim/hunter/kill_shot.go
ToxicKevinFerm: Fixed

sim/hunter/marksmanship/aimed_shot.go
ToxicKevinFerm: Fixed

sim/hunter/marksmanship/chimera_shot.go
ToxicKevinFerm: Removed. Like I wrote in the comment below, I don't think Chimera has the effects in cata anymore. Adjusted it.

sim/hunter/marksmanship/marksmanship.go
ToxicKevinFerm: Yeah, I think that was just an inheritance from wotlk. Cata AiS doesnt have a cd, fixed

sim/hunter/rapid_fire.go
ToxicKevinFerm: Check my implementation and see how it looks

sim/hunter/survival/black_arrow.go
ToxicKevinFerm: Fixed

sim/hunter/survival/explosive_shot.go
ToxicKevinFerm: Fixed

sim/hunter/survival/explosive_shot.go
ToxicKevinFerm: Yeah, I wasn't sure if I needed to do something special here as Explosive Shot ticks once on application always, and with the new rollover, it ticks on application and then the duration is still increased.

ui/hunter/beast_mastery/sim.ts
ToxicKevinFerm: Since it's a hunter specific debuff, would I have to define it in debuffs.go still, or can I just apply the debuff sort of like CoE using the psuedostat `BonusRangedAttackPowerTaken` at the OnSpellHitDealt in a Hunters mark spell?

sim/core/character.go
ToxicKevinFerm: I guess it doesn't. What do you think about my solution in the latest commit?

sim/hunter/arcane_shot.go
ToxicKevinFerm: Seems it should be normalized

sim/hunter/kill_command.go
ToxicKevinFerm: Made a change here, but the killing streak aura might need changing either way

ui/hunter/beast_mastery/sim.ts
rosenrusinov: Hunters mark should still be implemented in debuffs.go and exposed in the debuffs section because other sims want to be able to activate it (unless the spell is changed in cata to have no effect for other classes).

What i meant here is that instead of controlling if the aura is active or not with a player option for the hunter sim specifically you should define a hunters mark spell that just applies the debuff and users can do it from APL

sim/core/focus.go
```diff
@@ -29,38 +29,35 @@ type focusBar struct {
...
+		focusRefundMetrics:  unit.NewFocusMetrics(ActionID{OtherID: proto.OtherAction_OtherActionRefund}),
 	}
 }
 
 // Computes the focus thresholds.
-func (fb *focusBar) setupFocusThresholds() {
-	if fb.unit == nil {
-		return
-	}
+func (unit *Unit) SetupFocusThresholds() {
```
rosenrusinov, line 56: why is this a Unit function now? To me it made sense more to be a function of the focus bar as you need the focus bar for this function to do anything

ToxicKevinFerm: Because I couldn't access the focusBar anywhere outside, except where it was. Maybe I was wrong, but whatever go autocomplete thought `hunter.unit.focusBar` was not a thing I could do.

sim/hunter/serpent_sting.go
```diff
@@ -25,15 +28,12 @@ func (hunter *Hunter) registerSerpentStingSpell() {
 			},
 			IgnoreHaste: true, // Hunter GCD is locked at 1.5s
 		},
-
-		// Need to specially apply LethalShots here, because this spell uses an empty proc mask
-		//TOdo: add glyph of serpent sting modifier *6percent crit chance
-		BonusCritRating: 1 * (core.CritRatingPerCritChance * float64(hunter.Talents.ImprovedSerpentSting)) + core.TernaryFloat64(hunter.HasPrimeGlyph(proto.HunterPrimeGlyph_GlyphOfSerpentSting), 0.06, 0),
+		BonusCritRating: impSSCritChance + impSSCritChance + core.TernaryFloat64(hunter.HasPrimeGlyph(proto.HunterPrimeGlyph_GlyphOfSerpentSting), 6, 0)*core.CritRatingPerCritChance,
```
rosenrusinov, line 31: impSSCritChance is added twice

ToxicKevinFerm: Fixed

sim/hunter/survival/black_arrow.go
rosenrusinov: I've been reading a lot about dots but I see conflicting info from time to time. Its exactly Cynwises blog that I read that crit snapshotting is no longer a thing but maybe I read something wrong. I will have to keep on researching around the new mechanics as I want to tackle them very soon

sim/hunter/talents.go
```diff
@@ -1,870 +1,822 @@
...
+	if hunter.Pet != nil {
+		hunter.applyFrenzy()
+		hunter.registerBestialWrathCD()
+		// Todo: BM stuff
+		hunter.Pet.ApplyTalents()
+	}
+
+	if hunter.Talents.Pathing > 0 {
+		bonus := 0.01 * float64(hunter.Talents.Pathing)
+		hunter.MultiplyCastSpeed(bonus) //Todo: Should this be attackspeed?
```
rosenrusinov, line 21: yes to your comment. This should call `hunter.MultiplyRangedAttackSpeed` if it works only on ranged or `hunter.MultiplyAttackSpeed` if its supposed to mod both melee and ranged

rosenrusinov: Also looking at the talent its 0.03 per point not 0.01

sim/hunter/talents.go
```diff
@@ -1,870 +1,822 @@
...
+	}
+
+	if hunter.Talents.HunterVsWild > 0 {
+		bonus := 0.5 * float64(hunter.Talents.HunterVsWild)
```
rosenrusinov, line 25: 0.05 instead of 0.5. You're currently giving 50% bonus per point

sim/hunter/talents.go
```diff
@@ -1,870 +1,822 @@
...
+		hunter.MultiplyStat(stats.Stamina, 1+bonus)
+	}
+
+	if hunter.Talents.HuntingParty {
+		agiBonus := 0.01
```
rosenrusinov, line 30: 0.02

sim/hunter/survival/black_arrow.go
ToxicKevinFerm: The difference seems to be is that it always snapshots on application **and** refresh. So you can never carry a snapshotted dot over a whole fight. It might be then if you have a super snapshot on a dot, that you don't want to refresh it in the rollover period so you can get the big last tick.

ui/hunter/beast_mastery/sim.ts
ToxicKevinFerm: See my implementation

sim/hunter/survival/explosive_shot.go
ToxicKevinFerm: Fixed it yesterday to just apply and tickonce

sim/hunter/survival/black_arrow.go
rosenrusinov: Ah yes that would be it. It still snapshots but rollover now properly update everything so you cant just carry a very buffed dot forever

sim/core/focus.go
rosenrusinov: Ahh yeah that is true hmm. Now i'm conflicted about the first review I did. Do we want to keep the setup of the thresholds in the core so we dont have to expose the focusBar but then we have a bit of class code in the core. Which of the 2 is the lesser evil. Right now you have a method that you can call on any unit that works on focus bar but not every unit has a focus bar. Maybe its better if you go back to seting up the thresholds from core.

.gitignore
```diff
@@ -4,7 +4,8 @@ node_modules
 ui/core/index.ts
 ui/*/*/index.html
 .dirstamp
-
+.vscode/launch.json
+.vscode/settings.json
```
rosenrusinov, line 8: the settings json is actually our formatting setup so I dont think you want it in the gitignore because we want all our developers to have the same auto farm settings

rosenrusinov: Also you should rever the deletion of the settings.json for sure

sim/core/focus.go
```diff
@@ -48,23 +124,57 @@ func (fb *focusBar) CurrentFocus() float64 {
...
+	ticksPerSecond := float64(time.Second) / float64(FocusTickDuration)
+	hastePercent := fb.unit.RangedSwingSpeed()
+	tick := fb.baseFocusPerSecond * hastePercent / ticksPerSecond
+	return tick
+}
+
+func (fb *focusBar) onFocusGain(sim *Simulation, crossedThreshold bool) {
+	if sim.CurrentTime < 0 {
+		return
+	}
```
rosenrusinov, line 150: Does this make sense? If you have this check here any prepull actions that cost focus will leave you with no focus regen until the combat starts

.gitignore
ToxicKevinFerm: Yeah its just that the settings.json doesnt have settings for auto formatting go, so it was messing with my vscode

ToxicKevinFerm: Fixed

sim/hunter/cobra_shot.go
```diff
@@ -0,0 +1,53 @@
...
+			focus := 9.0
+			if hunter.Talents.Termination != 0 && sim.IsExecutePhase25() {
+				focus = float64(hunter.Talents.Termination) * 3
+			}
+			hunter.AddFocus(sim, focus, csMetrics)
+			result := spell.CalcDamage(sim, target, baseDamage, spell.OutcomeRangedHitAndCrit)
+			if hunter.SerpentSting.Dot(target).IsActive() {
+				hunter.SerpentSting.Dot(target).Rollover(sim)
+			}
+			spell.DealDamage(sim, result)
```
rosenrusinov, line 50: Is there a reason to store the result in a reference if you're not using it for anything? is the dot supposed to apply before the damage is dealt or something like that?

sim/hunter/hunter.go
rosenrusinov: Ah this then could just be moved to the core in character if its not hunter related but a shorthand for getting the current total points

sim/hunter/hunter.go
```diff
@@ -79,53 +90,55 @@ func (hunter *Hunter) GetHunter() *Hunter {
...
-	// if hunter.Options.UseHuntersMark {
-	// 	hunter.RegisterPrepullAction(0, func(sim *core.Simulation) {
-	// 		huntersMarkAura := core.HuntersMarkAura(hunter.CurrentTarget, hunter.Talents.ImprovedHuntersMark, hunter.HasMajorGlyph(proto.HunterMajorGlyph_GlyphOfHuntersMark))
-	// 		huntersMarkAura.Activate(sim)
-	// 	})
-	// }
+	hunter.AutoAttacks.MHConfig().CritMultiplier = hunter.CritMultiplier(false, false, false)
+	hunter.AutoAttacks.OHConfig().CritMultiplier = hunter.CritMultiplier(false, false, false)
+	hunter.AutoAttacks.RangedConfig().CritMultiplier = hunter.CritMultiplier(false, false, false)
+	hunter.FireTrapTimer = hunter.NewTimer()
```
rosenrusinov, line 123: this fire trap timer does not seem shared so it can be removed from here and just created in the fire trap?

sim/hunter/hunters_mark.go
```diff
@@ -0,0 +1,28 @@
...
+		Cast: core.CastConfig{
+			DefaultCast: core.Cast{
+				GCD: core.GCDDefault,
+			},
+			IgnoreHaste: true,
+		},
+
+		ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
+			hm := core.HuntersMarkAura(target)
+			hm.Activate(sim)
```
rosenrusinov, line 25: You are creating a new aura everytime you cast this spell.

What you shoould do is create an enemy aura array in the registration then in the cast activate the proper aura based on your target. 

If hunters mark can be active only on 1 target at the same time you still want to support to be able to activate it on any of the targets you have and make sure to deactivate it for the old target which can be done by just deactivating them all before activating it for the target (check NewEnemyAuraArray and how we use it around other debuffs)

sim/hunter/hunter.go
ToxicKevinFerm: Black Arrow uses it

sim/hunter/serpent_sting.go
```diff
@@ -0,0 +1,76 @@
...
+			NumberOfTicks: 5,
+			TickLength:    time.Second * 3,
+
+			OnSnapshot: func(sim *core.Simulation, target *core.Unit, dot *core.Dot, isRollover bool) {
+				dot.SnapshotBaseDamage = (460 + 0.40*dot.Spell.RangedAttackPower(target)) / 5
+				if !isRollover {
+					attackTable := dot.Spell.Unit.AttackTables[target.UnitIndex]
+					dot.SnapshotCritChance = dot.Spell.PhysicalCritChance(attackTable)
+					dot.SnapshotAttackerMultiplier = dot.Spell.AttackerDamageMultiplier(attackTable)
+				}
```
rosenrusinov, line 60: As we discussed in an earlier comment I think it no longer matters if its rollover or not snapshots are always recalculated

sim/hunter/steady_shot.go
```diff
@@ -0,0 +1,61 @@
...
+			baseDamage := hunter.AutoAttacks.Ranged().CalculateNormalizedWeaponDamage(sim, spell.RangedAttackPower(target)) + (280 + spell.RangedAttackPower(target)*0.021)
+			focus := 9.0
+			if hunter.Talents.Termination != 0 && sim.IsExecutePhase25() {
+				focus = float64(hunter.Talents.Termination) * 3
+			}
+
+			if hunter.Talents.MasterMarksman != 0 {
+				procChance := float64(hunter.Talents.MasterMarksman) * 0.2
+				if sim.Proc(procChance, "Master Marksman Proc") && !hunter.MasterMarksmanCounterAura.IsActive() {
+					hunter.MasterMarksmanCounterAura.Activate(sim)
```
rosenrusinov, line 53: I think you need to call aura.AddStack as well. I'm not sure if it is done automatically except for the 1st stack

sim/hunter/survival/black_arrow.go
```diff
@@ -0,0 +1,72 @@
...
+				//  66.5% RAP + 2849 (total damage) - changed 6/28 in 4.2 (based off spell crit multiplier, modified by toxicology)
+				// https://wago.tools/db2/SpellEffect?build=4.4.0.53750&filter[SpellID]=exact%3A3674&page=1
+				baseDamage := 2849.0
+				rap := dot.Spell.RangedAttackPower(target)
+				percentageOfRAP := 0.665
+
+				// SnapshotBaseDamage calculation for the DoT, divided by 10 to spread across all ticks
+				dot.SnapshotBaseDamage = baseDamage + (percentageOfRAP*rap)/10
+
+				dot.SnapshotAttackerMultiplier = dot.Spell.AttackerDamageMultiplier(dot.Spell.Unit.AttackTables[target.UnitIndex])
```
rosenrusinov, line 57: bwt you also need to snapshot the crit chance in `dot.SnapshotCritChance` check wotlk dots with crits for a reference

sim/hunter/cobra_shot.go
ToxicKevinFerm: No reason, it's changed now.

sim/hunter/talents.go
```diff
@@ -1,870 +1,823 @@
...
-	// hunter.registerReadinessCD()
+	if hunter.Pet != nil {
+		hunter.applyFrenzy()
+		hunter.registerBestialWrathCD()
+		// Todo: BM stuff
+		hunter.Pet.ApplyTalents()
+	}
+
+	if hunter.Talents.Pathing > 0 {
+		bonus := 0.03 * float64(hunter.Talents.Pathing)
```
rosenrusinov, line 20: Im sorry i know i said this should be 0.03 but after looking at the talent again its actually 1% per point as you had it I dunno what i was seeing😅

sim/core/focus.go
ToxicKevinFerm: Aight, reverted

sim/hunter/talents.go
```diff
@@ -1,870 +1,823 @@
...
+				hunter.AimedShot.DefaultCast.CastTime = 0
+			}
+		},
+		OnExpire: func(aura *core.Aura, sim *core.Simulation) {
+			if hunter.AimedShot != nil {
+				hunter.AimedShot.CostMultiplier = 1
+				hunter.AimedShot.DefaultCast.CastTime = time.Second * 3
+			}
+		},
+		OnSpellHitDealt: func(aura *core.Aura, sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
```
rosenrusinov, line 76: I think its better if this event is OnCastComplete as thats when the aura should be consumed not when the damage is dealt, right?

sim/hunter/talents.go
```diff
@@ -1,870 +1,823 @@
...
+			}
+
+		},
+	})
+	hunter.MasterMarksmanCounterAura = hunter.RegisterAura(core.Aura{
+		Label:     "Master Marksman",
+		Duration:  time.Second * 30,
+		ActionID:  core.ActionID{SpellID: 34486},
+		MaxStacks: 4,
+		OnSpellHitDealt: func(aura *core.Aura, sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
```
rosenrusinov, line 90: same comment as above. Shouldnt this work with the OnCastComplete event instead?

sim/hunter/talents.go
```diff
@@ -1,870 +1,823 @@
...
+			if spell != hunter.SteadyShot {
+				return
+			}
+			if sim.Proc(procChance, "Master Marksman Proc") {
```
rosenrusinov, line 94: you can optimize this check by checking `if procChance == 1 || sim.Proc(...` this would compltely skip rolling the random as you have 100% chance to apply it anyway

sim/hunter/talents.go
```diff
@@ -1,870 +1,823 @@
...
+
+	actionID := core.ActionID{SpellID: 83490}
+
+	focusMetrics := hunter.NewFocusMetrics(actionID)
+	core.MakePermanent(hunter.RegisterAura(core.Aura{
+		Label: "Termination",
+		OnSpellHitDealt: func(aura *core.Aura, sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
+			if sim.IsExecutePhase25() &&
+				spell == hunter.SteadyShot ||
+				spell == hunter.CobraShot {
```
rosenrusinov, line 219: no need for this formating when the line is not long enough to overflow the normal screen size

sim/hunter/talents.go
```diff
@@ -1,870 +1,823 @@
...
+	if hunter.Talents.ImprovedSteadyShot == 0 {
+		return
+	}
+	hunter.ImprovedSteadyShotAura = hunter.RegisterAura(core.Aura{
+		Label:     "Improved Steady Shot",
+		ActionID:  core.ActionID{SpellID: 53221, Tag: 1},
+		Duration:  time.Second * 8,
+		MaxStacks: 1,
+		OnGain: func(aura *core.Aura, sim *core.Simulation) {
+			attackspeedMultiplier := 1 + (float64(hunter.Talents.ImprovedSteadyShot) * 0.05)
```
rosenrusinov, line 294: this can be calculated outside of the OnGain to save computation

sim/hunter/talents.go
```diff
@@ -1,870 +1,823 @@
...
+	if hunter.Talents.KillingStreak == 0 {
+		return
+	}
+	hunter.KillingStreakAura = hunter.RegisterAura(core.Aura{
+		Label:     "Killing Streak",
+		ActionID:  core.ActionID{SpellID: 82748},
+		Duration:  core.NeverExpires,
+		MaxStacks: 1,
+		OnGain: func(aura *core.Aura, sim *core.Simulation) {
+			hunter.KillCommand.DamageMultiplier /= 1 + (float64(hunter.Talents.KillingStreak) * 0.1)
```
rosenrusinov, line 332: / instead of *

sim/hunter/talents.go
```diff
@@ -1,870 +1,823 @@
...
+			hunter.KillCommand.ApplyCostModifiers(hunter.KillCommand.CurCast.Cost - (float64(hunter.Talents.KillingStreak) * 5))
+		},
+		OnExpire: func(aura *core.Aura, sim *core.Simulation) {
+			hunter.KillCommand.DamageMultiplier /= 1 + (float64(hunter.Talents.KillingStreak) * 0.1)
+			hunter.KillCommand.ApplyCostModifiers(hunter.KillCommand.CurCast.Cost + (float64(hunter.Talents.KillingStreak) * 5))
+		},
+		OnSpellHitDealt: func(aura *core.Aura, sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
+			if hunter.Talents.KillingStreak > 0 {
```
rosenrusinov, line 340: not needed if. You cant get to this if yo udont have points in killing streak as you already return earlier with the same check

sim/hunter/talents.go
```diff
@@ -1,870 +1,823 @@
...
+	})
+	hunter.KillingStreakCounterAura = hunter.RegisterAura(core.Aura{
+		Label:     "Killing Streak (KC Crit)",
+		Duration:  core.NeverExpires,
+		MaxStacks: 2,
+		OnReset: func(aura *core.Aura, sim *core.Simulation) {
+			aura.Activate(sim)
+		},
+		OnSpellHitDealt: func(aura *core.Aura, sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
+			if hunter.Talents.KillingStreak > 0 {
```
rosenrusinov, line 355: same as above

sim/hunter/talents.go
```diff
@@ -1,870 +1,823 @@
...
+	if hunter.Pet != nil {
+		hunter.applyFrenzy()
+		hunter.registerBestialWrathCD()
+		// Todo: BM stuff
+		hunter.Pet.ApplyTalents()
+	}
+
+	if hunter.Talents.Pathing > 0 {
+		bonus := 0.03 * float64(hunter.Talents.Pathing)
+		hunter.MultiplyCastSpeed(bonus)
```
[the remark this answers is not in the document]
ToxicKevinFerm: Cant use those functions in talent because they modify swing timers no? I looked for them, MultiplyCastSpeed was the most reasonable one I could find. I'll do `unit.PseudoStats.RangedSpeedMultiplier *= amount` instead but again, dunno if the right value is there when im trying to set it.

sim/hunter/talents.go
```diff
@@ -1,870 +1,823 @@
...
+		ActionID: actionID,
+		Duration: time.Second * 20,
+		OnReset: func(aura *core.Aura, sim *core.Simulation) {
+			aura.Activate(sim)
+		},
+		OnGain: func(aura *core.Aura, sim *core.Simulation) {
+			hunter.Pet.FrenzyStacksSnapshot = float64(hunter.Pet.FrenzyAura.GetStacks())
+			if hunter.Pet.FrenzyStacksSnapshot >= 1 {
+				hunter.Pet.FrenzyAura.Deactivate(sim)
+				hunter.Pet.AddFocus(sim, 4, nil)
```
rosenrusinov, line 430: missing metrics

sim/hunter/talents.go
```diff
@@ -1,870 +1,823 @@
...
+
+func (hunter *Hunter) applyGoForTheThroat() {
+	if hunter.Talents.GoForTheThroat == 0 {
+		return
+	}
+	if hunter.Pet == nil {
+		return
+	}
+
+	spellID := []int32{0, 34950, 34950}[hunter.Talents.GoForTheThroat]
```
rosenrusinov, line 585: you already return if go for the throat is 0 so this spell id is always 34950

sim/hunter/talents.go
```diff
@@ -1,870 +1,823 @@
...
+	procChance := float64(hunter.Talents.ThrillOfTheHunt) * 0.05
+	focusMetrics := hunter.NewFocusMetrics(core.ActionID{SpellID: 34499})
+
+	hunter.RegisterAura(core.Aura{
+		Label:    "Thrill of the Hunt",
+		Duration: core.NeverExpires,
+		OnReset: func(aura *core.Aura, sim *core.Simulation) {
+			aura.Activate(sim)
+		},
+		OnSpellHitDealt: func(aura *core.Aura, sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
```
rosenrusinov, line 683: OnCastComplete instead of on damage?

sim/hunter/talents.go
```diff
@@ -1,870 +1,823 @@
...
+
+	hunter.RegisterAura(core.Aura{
+		Label:    "Hunting Party",
+		Duration: core.NeverExpires,
+		OnReset: func(aura *core.Aura, sim *core.Simulation) {
+			aura.Activate(sim)
+		},
+		OnGain: func(aura *core.Aura, sim *core.Simulation) {
+			aura.Unit.MultiplyAttackSpeed(sim, 1.10)
+		},
```
rosenrusinov, line 747: Missing OnExpire which would casuse this aura to constantly stack 10% attack speed on every single iteration

sim/hunter/talents.go
```diff
@@ -1,870 +1,823 @@
...
+			}
+
+			// TODO: This is needed because there are edge cases where core doesn't re-use Rapid Fire.
+			// Fix core so this isn't necessary.
+			core.StartDelayedAction(sim, core.DelayedActionOptions{
+				DoAt: sim.CurrentTime + 1,
+				OnAction: func(_ *core.Simulation) {
+					hunter.UpdateMajorCooldowns()
+				},
+			})
```
rosenrusinov, line 805: I dont think this is needed anymore with APL

sim/hunter/talents.go
```diff
@@ -1,870 +1,823 @@
...
+			IgnoreHaste: true, // Hunter GCD is locked
+			CD: core.Cooldown{
+				Timer:    hunter.NewTimer(),
+				Duration: time.Minute * 3,
+			},
+		},
+		ExtraCastCondition: func(sim *core.Simulation, target *core.Unit) bool {
+			// Don't use if there are no cooldowns to reset.
+			return !hunter.RapidFire.IsReady(sim)
+		},
```
rosenrusinov, line 774: This should not be needed anymore with APL

sim/hunter/trap_launcher.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+func (hunter *Hunter) registerTrapLauncher() {
+	actionID := core.ActionID{SpellID: 42899}
+
+	hunter.TrapLauncherAura = hunter.RegisterAura(core.Aura{
+		Label:    "Trap Launcher",
+		ActionID: actionID,
+		Duration: time.Second * 15,
+		OnSpellHitDealt: func(aura *core.Aura, sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
```
rosenrusinov, line 17: OnCastComplete?

sim/hunter/steady_shot.go
rosenrusinov: ok maybe im missremembering if you see proper behaviour then its fine

sim/hunter/talents.go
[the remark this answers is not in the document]
rosenrusinov: yes, if this is done during initialization and not dyanimacly then its the pseudo stat you have mentioned didnt realize this is not dynamic aura bonus

sim/hunter/hunter.go
ToxicKevinFerm: Moved out to core

sim/hunter/kill_command.go
ToxicKevinFerm: Fixed

PR #18, opened 2024-03-16
Virtual Scroll on log tab

ToxicKevinFerm:
Don't know if we want to do something with this @kayla-glick but rendering x amount of thousand elementsfor the log result isn't very performant.
This is not perfect, but we keep only 50 elements at a time like this.


kayla-glick:
Virtual scroll would be great! But it looks like there's maybe some issues to iron out?

It looks like a lot of the styling is missing

Old
image

New
image

And I'm not sure if this is just an issue with sims not being implemented, but the log is cutting off very early
image

ToxicKevinFerm:
> Virtual scroll would be great! But it looks like there's maybe some issues to iron out?

Definitely has some issues. Just seeing if it would be worth spending more time fixing them. But it seems so, so I'll look at it :D

ToxicKevinFerm:
image

Fixed styling and added a basic filter, splits on space and takes quotation marks for more specific filtering. Also a scroll to top button.

PR #30, opened 2024-03-20
Basic vengeance implementation

Swillis57:
More prep work for warrior masteries this time for prot in particular, based on what I could find in simc/wiki pages.

As commented I'm unsure about the exact behavior around non-hit attacks as the information I could dig up was conflicting without any patch notes to mark if/when it changed (aside from a note in 5.4 about dodges and parries continuing to work as they used to, but there's not enough to say if that only became a thing in MoP).

This is all easily tunable later once we can collect more data though. For now this should be enough to let people implement their tank specs.

### DIFF
--- a/sim/core/vengeance.go
+++ b/sim/core/vengeance.go
@@ -49,7 +49,7 @@ func UpdateVengeance(sim *Simulation, character *Character, tracker *VengeanceTr
 	}
 
 	// TODO: verify that stam actually shows up in this formula, currently implemented as simcraft has it
-	apBonusMax := character.GetStat(stats.Stamina) + 0.1*character.GetStat(stats.Health)
+	apBonusMax := character.GetStat(stats.Stamina) + 0.1*character.MaxHealth()
 	tracker.apBonus = Clamp(tracker.apBonus, 0, apBonusMax)
 
 	tracker.recentMaxAPBonus = math.Max(tracker.apBonus, tracker.recentMaxAPBonus)
--- a/sim/core/vengeance.go
+++ b/sim/core/vengeance.go
@@ -34,7 +34,7 @@ func UpdateVengeance(sim *Simulation, character *Character, tracker *VengeanceTr
 
 		// Decay existing bonus by half of the rate
 		decay := VengeanceAPDecayRate / 2
-		tracker.apBonus *= (1 - decay)
+		tracker.apBonus -= (decay * tracker.apBonus)
 
 		// Add 5% of damage taken in the last 2 seconds
 		tracker.apBonus += 0.05 * tracker.eligibleDamage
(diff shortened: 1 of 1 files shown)


sim/core/vengeance.go
```diff
@@ -0,0 +1,104 @@
...
+		// damage if the tank has been hit in the last 2 seconds
+		baseAPBonus := tracker.eligibleDamage / 3.0
+		tracker.apBonus = math.Max(tracker.apBonus, baseAPBonus)
+	} else {
+		// No hits in the last 2 seconds - apply full decay
+		tracker.apBonus -= (VengeanceAPDecayRate * tracker.recentMaxAPBonus)
+	}
+
+	// TODO: verify that stam actually shows up in this formula, currently implemented as simcraft has it
+	apBonusMax := character.GetStat(stats.Stamina) + 0.1*character.GetStat(stats.Health)
```
rosenrusinov, line 52: I think we prefer if we do any health queries trough the HealthBar so change `character.GetStat(stats.Health)` to `character.MaxHealth()`

sim/core/vengeance.go
```diff
@@ -0,0 +1,104 @@
...
+
+	// Vengeance "ticks" every 2 seconds to update the AP buff
+	character.RegisterResetEffect(func(sim *Simulation) {
+		StartPeriodicAction(sim, PeriodicActionOptions{
+			Period: time.Second * 2,
+			OnAction: func(sim *Simulation) {
+				UpdateVengeance(sim, character, tracker, vengAura)
+			},
+		})
+	})
```
rosenrusinov, line 103: Make sure to test if this action is not stacking with every iteration as you are activating it on every reset. I dont remember if we automatically cancel all PAs at the end of a run (we should but leaving this comment as a sanity check)

rosenrusinov (commented):
Left some minor comments. Thanks for taking the time to add this. It would be very handy as I will be starting the work on DK soon and will need it for the Blood spec.

Also on another note: I see you are actively adding core functionality but I havent seen you in our discord. Can you make sure to join there as a lot of organization and communication happens mainly there :)

sim/core/vengeance.go
Swillis57: PAs do get cleared before each run here: https://github.com/wowsims/cata/blob/621230372254ac470d5df72537ce1d951ca80a7d/sim/core/sim.go#L377

This is also copying how a couple other similar effects were implemented in Wrath (warrior's Anger Management rage tick, priest's Rapture mana tick) so it seems it should still be okay.

sim/core/vengeance.go
Swillis57: Oh duh, that's what I get for writing code at 6am. Fixedd!

rosenrusinov (approved):

PR #32, opened 2024-03-21
Add all Potions, Flasks and Elixirs

ToxicKevinFerm:
Added them, removed some old stuff that has replacements, and updated presets to be able to compile

### DIFF
--- a/sim/core/consumes.go
+++ b/sim/core/consumes.go
@@ -3,7 +3,6 @@ package core
 import (
 	"time"
 
-	"github.com/wowsims/cata/sim/core"
 	"github.com/wowsims/cata/sim/core/proto"
 	"github.com/wowsims/cata/sim/core/stats"
 )
@@ -15,7 +14,7 @@ func applyConsumeEffects(agent Agent) {
 	if consumes == nil {
 		return
 	}
-	alchemyFlaskBonus := core.TernaryInt(character.HasProfession(proto.Profession_Alchemy), 80, 0)
+	alchemyFlaskBonus := TernaryFloat64(character.HasProfession(proto.Profession_Alchemy), 80, 0)
 	if consumes.Flask != proto.Flask_FlaskUnknown {
 		switch consumes.Flask {
 		case proto.Flask_FlaskOfTitanicStrength:
@@ -588,19 +587,20 @@ func makePotionActivationInternal(potionType proto.Potions, character *Character
 		healthMetrics := character.NewHealthMetrics(actionID)
 		return MajorCooldown{
 			Type: CooldownTypeSurvival,
+			ShouldActivate: func(sim *Simulation, character *Character) bool {
+				// Only pop if we have less than the max mana provided by the potion minus 1mp5 tick.
+				totalRegen := character.ManaRegenPerSecondWhileCasting() * 5
+				manaGain := 11000.0
+				if alchStoneEquipped && potionType == proto.Potions_MythicalManaPotion {
+					manaGain *= 1.4
+				}
+				return character.MaxMana()-(character.CurrentMana()+totalRegen) >= manaGain
+			},
 			Spell: character.GetOrRegisterSpell(SpellConfig{
 				ActionID: actionID,
 				Flags:    SpellFlagNoOnCastComplete,
 				Cast:     potionCast,
-				ShouldActivate: func(sim *Simulation, character *Character) bool {
-					// Only pop if we have less than the max mana provided by the potion minus 1mp5 tick.
-					totalRegen := character.ManaRegenPerSecondWhileCasting() * 5
-					manaGain := 10750.0
-					if alchStoneEquipped && potionType == proto.Potions_MythicalManaPotion {
-						manaGain *= 1.4
-					}
-					return character.MaxMana()-(character.CurrentMana()+totalRegen) >= manaGain
-				},
+
 				ApplyEffects: func(sim *Simulation, _ *Unit, _ *Spell) {
 					resourceGain := sim.RollWithLabel(9000, 11000, "MightyRejuvPotion") // Todo: Does it roll once or twice?
 
--- a/sim/core/consumes.go
+++ b/sim/core/consumes.go
@@ -592,6 +592,15 @@ func makePotionActivationInternal(potionType proto.Potions, character *Character
 				ActionID: actionID,
 				Flags:    SpellFlagNoOnCastComplete,
 				Cast:     potionCast,
+				ShouldActivate: func(sim *Simulation, character *Character) bool {
+					// Only pop if we have less than the max mana provided by the potion minus 1mp5 tick.
+					totalRegen := character.ManaRegenPerSecondWhileCasting() * 5
+					manaGain := 10750.0
+					if alchStoneEquipped && potionType == proto.Potions_MythicalManaPotion {
+						manaGain *= 1.4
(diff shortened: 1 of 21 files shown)


sim/core/consumes.go
```diff
@@ -17,6 +17,37 @@ func applyConsumeEffects(agent Agent) {
...
 		switch consumes.Flask {
+		case proto.Flask_FlaskOfTitanicStrength:
+			character.AddStats(stats.Stats{
+				stats.Strength: 300,
+			})
+			if character.HasProfession(proto.Profession_Alchemy) {
+				character.AddStats(stats.Stats{
+					stats.Strength: 80,
+				})
+			}
```
rosenrusinov, line 28: Its prefered if we do a single AddStats instead of 2 so better to change all the alchemy bonuses to be added to a variable and then a single add stats with the total ammount

sim/core/consumes.go
```diff
@@ -640,28 +600,39 @@ func makePotionActivationInternal(potionType proto.Potions, character *Character
...
+		// No specific aura stats provided; adjust as needed
 		manaMetrics := character.NewManaMetrics(actionID)
+		healthMetrics := character.NewHealthMetrics(actionID)
 		return MajorCooldown{
-			Type: CooldownTypeMana,
-			ShouldActivate: func(sim *Simulation, character *Character) bool {
-				// Only pop if we have less than the max mana provided by the potion minus 1mp5 tick.
-				totalRegen := character.ManaRegenPerSecondWhileCasting() * 5
-				return character.MaxMana()-(character.CurrentMana()+totalRegen) >= 3000
-			},
```
rosenrusinov, line 653: I think you do want the rejuv potion to still have the same ShouldActivate or you're gonna be wasting the mana gain from i

ui/core/components/inputs/consumables.ts
```diff
@@ -353,28 +351,13 @@ export const GiftOfArthas = {
...
-export const ElixirOfMajorMageblood = {
-	actionId: ActionId.fromItemId(22840),
-	value: GuardianElixir.ElixirOfMajorMageblood,
-};
 
 export const GUARDIAN_ELIXIRS_CONFIG = [
+	{ config: ElixirOfDeepEarth, stats: [Stat.StatArmor] },
+	{
+		config: PrismaticElixir,
+		stats: [Stat.StatArcaneResistance, Stat.StatFireResistance, Stat.StatFrostResistance, Stat.StatNatureResistance, Stat.StatShadowResistance],
```
rosenrusinov, line 359: Prismatic should give all stats as well right? If so it should then be filtered on those primary stats as well. I dont think a single sim has resistances as an EP value so this poition will not show up on any UI

ui/core/components/inputs/consumables.ts
```diff
@@ -519,81 +502,69 @@ export const PetScrollOfStrengthV = makeBooleanConsumeInput({
...
-	{ config: RunicManaInjector, stats: [Stat.StatIntellect] },
-	{ config: IndestructiblePotion, stats: [Stat.StatArmor] },
-	{ config: InsaneStrengthPotion, stats: [Stat.StatStrength] },
-	{ config: HeroicPotion, stats: [Stat.StatStamina] },
+	{ config: GolembloodPotion, stats: [Stat.StatStrength] },
+	{ config: PotionOfTheTolvir, stats: [Stat.StatAgility] },
+	{ config: PotionOfConcentration, stats: [Stat.StatMana] },
+	{ config: VolcanicPotion, stats: [Stat.StatIntellect] },
+	{ config: EarthenPotion, stats: [Stat.StatArmor] },
+	{ config: MightyRejuvenationPotion, stats: [Stat.StatMana, Stat.StatHealth] },
```
rosenrusinov, line 557: I think we may have to add Mana as an EP stat to the sims that have mana as its no longer connected to intellect right?

ui/core/components/inputs/consumables.ts
```diff
@@ -519,81 +502,69 @@ export const PetScrollOfStrengthV = makeBooleanConsumeInput({
...
-	},
 ] as ConsumableStatOption<Potions>[];
 
 export const PRE_POTIONS_CONFIG = [
-	{ config: IndestructiblePotion, stats: [Stat.StatArmor] },
-	{ config: InsaneStrengthPotion, stats: [Stat.StatStrength] },
-	{ config: HeroicPotion, stats: [Stat.StatStamina] },
+	{ config: GolembloodPotion, stats: [Stat.StatStrength] },
+	{ config: PotionOfTheTolvir, stats: [Stat.StatAgility] },
+	{ config: VolcanicPotion, stats: [Stat.StatIntellect] },
```
rosenrusinov, line 566: Why are not all potions added as an option in the pre-pot? I dont think we want to assume what the user wants to do. If they want to prepot a potion that doesnt make sense for us they should still be able to do so

sim/core/consumes.go
ToxicKevinFerm: I figured that we should allow this to be handled in the APL? In a tank sim for a prot pala or whatever, it might not be used primarily for mana at all

ui/core/components/inputs/consumables.ts
ToxicKevinFerm: https://www.wowhead.com/cata/spell=80492/prismatic-elixir just the resistances

ui/core/components/inputs/consumables.ts
ToxicKevinFerm: It's only changed in mop. Cata still increases mana by intellect, I dont think that has changed going from wotlk to cata

ui/core/components/inputs/consumables.ts
ToxicKevinFerm: The only potion not in that list that makes sense to add is Earthen Potion, everything else is mana or health gain. But in that case, why even have separate configs?
I can just do
`export const PRE_POTIONS_CONFIG = [
	...POTIONS_CONFIG
] as ConsumableStatOption<Potions>[]; or refactor both to use the same config.
`

sim/core/consumes.go
ToxicKevinFerm: Fixed for the new flasks, let the old ones be there because i dont know the logic for them, and they seems to have different values for each one

ui/core/components/inputs/consumables.ts
ToxicKevinFerm: I just added the earthen pot for now, I dont think it makes sense at all to have mana and health pots as prepots

ui/core/components/inputs/consumables.ts
rosenrusinov: In that case the stat that this potion is tied to should be Intellect as we're already including it in sims EP values

ui/core/components/inputs/consumables.ts
rosenrusinov: Yeah. Mana/Health is not needed indeed and its better if they dont clutter the ui for the prepot then

sim/core/consumes.go
ToxicKevinFerm: Ah okay

ui/core/components/inputs/consumables.ts
ToxicKevinFerm: Fixed

rosenrusinov (approved):

PR #34, opened 2024-03-23
Port SoD movement and add new Is Moving APL Value check

ToxicKevinFerm:
Ported the SoD implementation 1:1, but also added a "Is Moving" APL value because hunters and other classes that have conditions where they can cast on the move might want to implement those.

Take Hunter as an example, it'd be interesting to sim a fight with intermittent movement and when it's worth swapping to Fox to cast something along with focus management. Can just remove it from the list if we dont want it though.

### DIFF
--- a/proto/apl.proto
+++ b/proto/apl.proto
@@ -41,7 +41,7 @@ message APLListItem {
     APLAction action = 3; // The action to be performed.
 }
 
-// NextIndex: 20
+// NextIndex: 21
 message APLAction {
     APLValue condition = 1; // If set, action will only execute if value is true or != 0.
 
@@ -69,16 +69,17 @@ message APLAction {
         APLActionCancelAura cancel_aura = 10;
         APLActionTriggerICD trigger_icd = 11;
         APLActionItemSwap item_swap = 17;
+		APLActionMove move = 18;
 
         // Class or Spec-specific actions
-        APLActionCatOptimalRotationAction cat_optimal_rotation_action = 18;
+        APLActionCatOptimalRotationAction cat_optimal_rotation_action = 19;
 
         // Internal use only, not exposed in UI.
-        APLActionCustomRotation custom_rotation = 19;
+        APLActionCustomRotation custom_rotation = 20;
     }
 }
 
-// NextIndex: 67
+// NextIndex: 68
 message APLValue {
     oneof value {
         // Operators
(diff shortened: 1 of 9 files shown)


rosenrusinov:
You seem to have missed some code that is in the cast.go that prevents hardcasts from starting while moving

rosenrusinov:
Also I actually wanted to refactor this in SoD as well as I'm not entirely happy with how I implemented some stuff (cant move in fractions and the movement tick is static instead of dynamic based on move ttime, also my estimation of move speed was way off and since then I know players move at 7y/s)

rosenrusinov:
Also I wanted to add a new pseudoStat for move speed bonuses as that existst and does make a difference ingame

ToxicKevinFerm:
> Also I wanted to add a new pseudoStat for move speed bonuses as that existst and does make a difference ingame

Yeah, especially in Cataclysm all of the bis feet enchants are with movement speed as rocket boots start failing in raids.
I'll leave this branch if you want to take it up at some point for the refactor.

InDebt:
What's the state on this @ToxicKevinFerm you're working on this PR still? Got lots of queries already about supporting Movement triggers in APL as shadow i.E. has talents that have increased proc chance while moving

rosenrusinov:
This is actually on me. I will look into it

InDebt:
Yeah no pressure. Was just wondering as I saw this pending for some time. Can also take this up as I have public holliday on wednesday.

rosenrusinov:
Yeah if you get to this before me be my guest to do it xD
What I wanted to do:
- In the current move to yards functionality we should be able to move to any fraction of yards. Currently it bugs. It also needs a refactor in the speed. I now know that characters move at 7y/s.
- This PR is missing from the original implementation some extra code in cast.go and attack.go for handling melee/ranged attacks activation/deactivation based on range and hardcasting based on movement state. This needs to be reimplemented and maybe changed a bit for cata as hunter can autoshoot while moving if i remember correctly.
- I wanted to add another apl action for movement which is based on time and does not change the distance to target but just sets the movement state for a certain time. This would be used to simulate movement when boss phases require it or for activating effects based on movement like the priest ones

@InDebt

ToxicKevinFerm:
Maybe we'd want to add min and max range on spells too? Depending on boss AI, a boss might move out of range for a period of time for example

PR #35, opened 2024-03-23
Add Hunter travel time on spells

ToxicKevinFerm:


rosenrusinov (commented):
Everything looks good except multi shot which is now calculating its damage when the missiles lands instead of when fired

rosenrusinov (commented):
You still need to move the CalcDamage call before the wait travel time and store the results in an array. Then apply the results in the land

rosenrusinov (approved):

PR #40, opened 2024-03-28
[Enchants] Adds enchants

ToxicKevinFerm:
Todo: Fix correct Item Quality. Use items where applicable.
Also todo: implement the spell effect enchants.
I believe I have all of the enchants. About 114, with more or less correct stats. 
I can't get them to show up in the UI though. They're only in the leftover db and idk what happens there, but I dont see them in the db.json when the ui loads it.


rosenrusinov:
Can be closed due to #54

PR #42, opened 2024-03-30
[Rogue] Some small fixes after Beta testing, + Fix for DB Generator

TheBackstabi:
Makes some abilities function closer to their implementation on Beta, and adds a nil check to Assassin's Resolve to fix DB generation.


sim/rogue/combat/revealing_strike.go
```diff
@@ -24,18 +25,23 @@ func (comRogue *CombatRogue) registerRevealingStrike() {
...
+					comRogue.Rupture.DamageMultiplier *= multiplier
+					isApplied = true
+				}
+			},
+			OnExpire: func(aura *core.Aura, sim *core.Simulation) {
+				comRogue.Eviscerate.DamageMultiplier /= multiplier
+				comRogue.Envenom.DamageMultiplier /= multiplier
+				comRogue.Rupture.DamageMultiplier /= multiplier
+				isApplied = false
+				aura.Deactivate(sim)
```
kayla-glick, line 41: Revealing Strike is applied to each target as a debuff right? This implementation flat buffs the rogue's next strike regardless of target. I think the right way to do this would be to register a debuff on the target, then in the various finisher calculations, check if the buff is active and apply the multiplier

PR #45, opened 2024-04-04
[Shadow] Baseline Implementation

InDebt:
Cheers!

I have tried to implement the baseline for the shadow spec module. I have never worked with this code base nor did I ever code in Go before, so feel free to point out anything! I feel after reviewing the code that I kind off stuck to what I know best and that's the way aura mods and spell effects are handled on the p-server side, as I developed there for serveral years. I recon that that's somewhat different as responsiblities seem to be mixed across the code base as to where dmg mods or other effects are applied. i.E. PseudoStats vs direct calculation within the spells. So let me know if that has to be changed.

## Included
* Mind Quickening Raid Buff
* All relevant stat conversions
* All shadow relevant talents
* All major shadow glyphs
* All relevant shadow spells (Mind Spike not yet but it seems to be not really used looking at shadow guides)

## Open
* DoT System needs to be updated to cata
* Mana regen system is still WoTLK MP5 based
* Raid Buffs are still primarily WoLTK
* Provide proper default item set
* Provide proper default APL
* Verfiy implementation vs lvl 85 char - only did some sanity checks vs my lvel 80 char


sim/core/buffs.go
```diff
@@ -1524,3 +1529,49 @@ func FocusMagicAura(caster *Unit, target *Unit) (*Aura, *Aura) {
...
+		Duration:   NeverExpires,
+		BuildPhase: CharacterBuildPhaseBuffs,
+		OnReset: func(aura *Aura, sim *Simulation) {
+			aura.Activate(sim)
+		},
+	})
+
+	SpellHasteBonusEffect(aura, 0.05)
+	return aura
+}
```
kayla-glick, line 1577: Would you mind also adding Moonkin aura here? I saw you included it in the UI so we should include it in the exclusive effects

sim/priest/devouring_plague.go
```diff
@@ -0,0 +1,63 @@
...
+				dot.SnapshotAttackerMultiplier = dot.Spell.AttackerDamageMultiplier(dot.Spell.Unit.AttackTables[target.UnitIndex])
+			},
+			OnTick: func(sim *core.Simulation, target *core.Unit, dot *core.Dot) {
+				dot.CalcAndDealPeriodicSnapshotDamage(sim, target, dot.OutcomeSnapshotCrit)
+			},
+		},
+
+		ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
+			result := spell.CalcOutcome(sim, target, spell.OutcomeMagicHit)
+			spell.DealOutcome(sim, result)
```
kayla-glick, line 47: For this since you don't have any separate logic here you could just do `spell.CalcAndDealOutcome`

sim/core/buffs.go
InDebt: Should be available now. Removed it from the presets where possible and another haste buff was also present. Would love to add the other raid buffs as well but that will happen on another branch.
Gotta see how I can do that without running into many merge conflicts with all the other branches like warrior or shaman also implementing some buffs already 😁

sim/priest/priest.go
```diff
@@ -206,3 +257,190 @@ func New(char *core.Character, selfBuffs SelfBuffs, talents string) *Priest {
...
+	School     core.SpellSchool
+	BaseValue  T
+
+	// dynamic evaluation will be added to the given BaseValue
+	DynamicValue func(*Priest) T
+
+	// Stacks * (BaseValue + DynamicValue)
+	Stacks  int32
+	SpellID int32
+}
```
kayla-glick, line 350: The concept of tracking these aura mods and the various helpers makes sense to me but I wonder if this should be something we make general-purpose for all classes rather than just shadow-specific?

InDebt: Yeah it's how at least p-servers handle those mods internally.

https://wago.tools/db2/SpellClassOptions?build=4.4.0.54027&filter[SpellID]=exact%3A2944&page=1 - Class mask for Devouring 
https://wago.tools/db2/SpellEffect?build=4.4.0.54027&filter[SpellID]=63626&page=1 - Improved Devouring Plague Talent has the matching class mask - thus the dummy (usually the backing lua script) is only applied to that ability.

For now I've 'hacked' the damage mod applications by putting a wrapper around the spell config and changing the respective fields of the spell before calling the tick / apply methods. So it's not realy working nicely with the PseudoStats currently 😁

rosenrusinov: I don't like this implementation at all. It completely ignores what the core supports and implements the same thing but specific to the priest. It also makes unneccessery calculations during ApplyEffects which can be done once during initialization using the spell configs damage multipliers. 
I can see why this is how you want to do it on an actual server where those things need to be calculated dynamically but there is no reason to do it like this in the sim. We want all static modifiers to be calculated once on sim creation and not done on every application when you run thousends of iterations.

InDebt: Totally get that. Was easier for me to get a handle of things as keeping track of everything for every spell was a bit overwhelming at first, but after I have implemented everything for shadow now I can refactor everything into the respective spells / PseudoStats. Will push a commit later for that!

rosenrusinov: We can still have helper methods for building the damage mods in a way similiar to what you have done as long as its done outside of the ApplyEffects. Also as @kayla-glick was saying we may want to consider making this a part of core if people favour building the mods in such a way (Instead of what we currently do with having it explicitly build in their registration methods as I also agree with your opinion of this being overwhelming when you do the initial implementation)

sim/core/spell_mod.go
```diff
@@ -0,0 +1,287 @@
...
+
+	unit.OnSpellRegistered(func(spell *Spell) {
+		if shouldApply(spell, mod) {
+			mod.AffectedSpells = append(mod.AffectedSpells, spell)
+
+			if mod.IsActive {
+				mod.Apply(mod, spell)
+			}
+		}
+	})
```
rosenrusinov, line 75: OnSpellRegistered already calls the handler for each spell that is already registered so you dont need to do the above iteration over the spellbook

sim/core/spell_mod.go
```diff
@@ -0,0 +1,287 @@
...
+
+	return mod
+}
+
+func (unit *Unit) AddStaticMod(config SpellModConfig) {
+	mod := buildMod(unit, config)
+	unit.ActiveSpellMods = append(unit.ActiveSpellMods, mod)
```
rosenrusinov, line 82: where is this `unit.ActiveSpellMods` used? I dont see it anywhere so wondering why its needed. Also shouldnt the Activate/Deactivate call be responsible for adding the mod to that list anyway?

sim/core/spell_mod.go
InDebt: Oh sweet I missed that part. That's of course redundant then.

sim/core/spell_mod.go
InDebt: It felt odd at first for the static's mod single point of reference to be through the callback handler of 'OnSpellRegistered'. Yet, you are right I never actually used it anywhere I believe. I'll clean that up

sim/priest/talents.go
```diff
@@ -60,9 +61,644 @@ func (priest *Priest) ApplyTalents() {
...
+		OnReset: func(aura *core.Aura, sim *core.Simulation) {
+			aura.Activate(sim)
+		},
+
+		OnSpellHitDealt: func(aura *core.Aura, sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
+			// check for critical mind flay tick
+			if result.Outcome.Matches(core.OutcomeCrit) && priest.MindFlayAPL.ClassSpellMask == int64(PriestSpellMindFlay) {
+
+				// reduce cooldown
+				remaining := max(0, time.Duration(*priest.Shadowfiend.CD.Timer)-time.Second*5)
```
rosenrusinov, line 623: `CD.TimeToReady(sim)` or `CD.ReadyAt()` depending on what you need (You can check the other available methods you should not need to cast this to anything)

sim/priest/talents.go
```diff
@@ -60,9 +61,644 @@ func (priest *Priest) ApplyTalents() {
...
+		ClassSpellMask:           int64(PriestSpellShadowyApparation),
+		DamageMultiplier:         1,
+		DamageMultiplierAdditive: 1,
+		CritMultiplier:           priest.DefaultSpellCritMultiplier(),
+		Cast: core.CastConfig{
+			DefaultCast: core.Cast{
+				Cost: 0,
+				GCD:  0,
+			},
+		},
```
rosenrusinov, line 651: empty castconfig should be not needed can just remove it

sim/priest/talents.go
```diff
@@ -60,9 +61,644 @@ func (priest *Priest) ApplyTalents() {
...
+				// spell and other modifiers are snapshotted when the apparations spawn
+				spell.DamageMultiplier = dmgMulti
+				spell.DamageMultiplierAdditive = dmgMultiAdd
+
+				result := spell.CalcDamage(sim, target, baseDamage, spell.OutcomeMagicHitAndCrit)
+				spell.DealDamage(sim, result)
+
+				// restore mods
+				spell.DamageMultiplier = oldMulti
+				spell.DamageMultiplierAdditive = oldAdd
```
rosenrusinov, line 679: Usually all Attacker(Caster) mods are applied on CastComplete (when the apparition is cast) and target mods are applied when the damage is done (when the apparition hits the target). You dont need to snapshot any modifiers manually instead you do:
```
result := spell.CalcDamage(sim, target, baseDamage, outcom...)
spell.WaitTravelTime(sim, func(sim *core.Simulation) {
    spell.DealDamage(sim, result)
}
```
what this does is calculate the damage (+caster modifiers on the cast complete) and then after travel time deal the damage(applying target modifiers in the dealing part)

rosenrusinov: Basically Caster mods are applied in the Calc part and target mods are applied in the Deal part

sim/priest/shadow/shadow.go
```diff
@@ -62,3 +73,125 @@ func (spriest *ShadowPriest) Initialize() {
...
+}
+
+func handleShadowOrbPower(spriest *ShadowPriest, sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
+	if !result.Landed() {
+		return
+	}
+
+	if spell == spriest.ShadowWordPain || spell.SpellID == spriest.MindFlayAPL.SpellID {
+		procChance := 0.1 + float64(spriest.Talents.HarnessedShadows)*0.04
+		if sim.RandomFloat("Shadow Orb Power") < procChance {
```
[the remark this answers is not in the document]
rosenrusinov: This should be applied to other places in this MR that `sim.RandomFloat` is used instead

sim/priest/talents.go
```diff
@@ -60,9 +61,644 @@ func (priest *Priest) ApplyTalents() {
...
+	if priest.Talents.ImprovedDevouringPlague == 0 {
+		return
+	}
+
+	// simple spell here as it does not use any dmg mods or calculations
+	impDPDamage := priest.Unit.RegisterSpell(core.SpellConfig{
+		ActionID:                 core.ActionID{SpellID: 63675},
+		SpellSchool:              core.SpellSchoolShadow,
+		ProcMask:                 core.ProcMaskProc,
+		Flags:                    core.SpellFlagIgnoreAttackerModifiers | core.SpellFlagIgnoreTargetModifiers | core.SpellFlagNoSpellMods,
```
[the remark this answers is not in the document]
InDebt: I wonder if I should add the new Flag to that combined modifier as well?

sim/priest/talents.go
```diff
@@ -60,9 +61,644 @@ func (priest *Priest) ApplyTalents() {
...
+		DamageMultiplierAdditive: 1,
+		ThreatMultiplier:         1,
+		CritMultiplier:           priest.DefaultSpellCritMultiplier(),
+		ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
+			dp := priest.DevouringPlague
+			dot := dp.Dot(target)
+
+			// No need to use snapshot dmg. It won't be initialized in time and we're in the same sim cycle
+			// so all mods and buffs will be the same
+			dmg := float64(dot.NumberOfTicks*int32(dp.ExpectedTickDamage(sim, target))*priest.Talents.ImprovedDevouringPlague) * 0.15
```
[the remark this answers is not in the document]
InDebt: Well snapshot aside, I thought it was exactly for this.

So to elaborate. I implemented Devouring Plague like this right now. As I need the spell hit for the imp talent to proc. However, I can obviously do `CalcOutcome`, apply the dot and then do `DealOutcome` so the snapshot is calculated before the spell hit happend.
```
	result := spell.CalcAndDealOutcome(sim, target, spell.OutcomeMagicHit)
	if result.Landed() {
		spell.SpellMetrics[target.UnitIndex].Hits--
		spell.Dot(target).Apply(sim)
	}
```

Improved Devouring plague deals 30% of the DoTs damage over time instantly when applied. So the thought was just using the existing 'ExpectedTickDamage' method and multiply that by the calculated Ticks to get the total damage.

sim/priest/talents.go
```diff
@@ -60,9 +61,644 @@ func (priest *Priest) ApplyTalents() {
...
+				damageTakenHandler(sim, result.Damage)
+			},
+
+			OnSpellHitDealt: func(aura *core.Aura, sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
+				// SW:D
+				if priest.ShadowWordDeath == spell && result.Landed() {
+					priest.AddMana(sim, priest.MaxMana()*0.05*float64(priest.Talents.Masochism), manaMetrics)
+					return
+				}
+			},
```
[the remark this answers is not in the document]
InDebt: If you say that's no problem. I know sims that just say - yo you're dead. Sim stopped, because there's no incoming heal.

sim/priest/talents.go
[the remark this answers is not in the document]
rosenrusinov: > So the thought was just using the existing 'ExpectedTickDamage' method and multiply that by the calculated Ticks to get the total damage.

Using `ExpectedTickDamage` would mean you run the whole damage pipeline twice for the same outcome. You will notice on other places that we do exactly what you propose where we apply the dots before dealing the damage.

sim/priest/talents.go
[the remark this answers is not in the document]
rosenrusinov: Our core will also just track that you have died and continue running normally so its no problem to do it. Another option is to just do a `CalcAndDealOutcome` on the priest which doesnt deal actual damage but still runs damage events

sim/priest/priest.go
```diff
@@ -206,3 +201,90 @@ func New(char *core.Character, selfBuffs SelfBuffs, talents string) *Priest {
...
+// Other flags
+type PriestSpellFlag uint64
+
+// Returns whether there is any overlap between the given masks.
+func (se PriestSpellFlag) Matches(other PriestSpellFlag) bool {
+	return (se & other) != 0
+}
+
+const (
+	PriestSpellFlagNone  PriestSpellFlag = 0
```
rosenrusinov, line 215: this should probably just be int64 if you're constantly casting it to that anyway

sim/priest/glyphs.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	"time"
+
+	"github.com/wowsims/cata/sim/core"
+	"github.com/wowsims/cata/sim/core/proto"
+)
+
+func (priest *Priest) ApplyGlyphs() {
+	// Glphys
+	if priest.HasGlyph(int32(proto.PriestPrimeGlyph_GlyphOfShadowWordPain)) {
```
rosenrusinov, line 12: btw I didnt notice the first time that you used the `HasGlyph` func. There should be explicit functions for checking `HasPrimeGlyph`, `HasMajorGlyph` etc.

sim/priest/talents.go
```diff
@@ -445,18 +402,15 @@ func (priest *Priest) applyImprovedDevouringPlague() {
...
 		CritMultiplier:           priest.DefaultSpellCritMultiplier(),
 		ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
-			dp := priest.DevouringPlague
-			dot := dp.Dot(target)
-
-			// No need to use snapshot dmg. It won't be initialized in time and we're in the same sim cycle
-			// so all mods and buffs will be the same
-			dmg := float64(dot.NumberOfTicks*int32(dp.ExpectedTickDamage(sim, target))*priest.Talents.ImprovedDevouringPlague) * 0.15
+			dot := priest.DevouringPlague.Dot(target)
+			dpTickDamage := dot.CalcSnapshotDamage(sim, target, dot.OutcomeExpectedMagicSnapshotCrit)
```
rosenrusinov, line 412: `dpTickDamage := dot.SnapshotBaseDamage` it should be already calculated. Also using the `OutcomeExpectedMagicSnapshotCrit` will bake crit inside the base damage then you apply another outcome that can crit as well making this double dip on crit which seems wrong.

Just using the dots snapshotted base damage should give you exactly that the expected non-crit damage per tick.

sim/priest/talents.go
```diff
@@ -445,18 +402,15 @@ func (priest *Priest) applyImprovedDevouringPlague() {
...
+			dmg := float64(dot.NumberOfTicks) * dpTickDamage.Damage * float64(priest.Talents.ImprovedDevouringPlague) * 0.15
 			spell.CalcAndDealDamage(sim, target, dmg, spell.OutcomeMagicCrit)
```
rosenrusinov, line 414: `OutcomeMagicCrit` is guaranteed critical. I think you want `OutcomeMagicHitAndCrit`

rosenrusinov: Nope my bad. The outcome you are using is either a crit or a hit, never a miss which is what you want

sim/priest/talents.go
```diff
@@ -60,9 +60,591 @@ func (priest *Priest) ApplyTalents() {
...
+
+	// Shadowform
+	if priest.Talents.Shadowform {
+		// no class restrictions
+		priest.AddStaticMod(core.SpellModConfig{
+			School:     core.SpellSchoolShadow,
+			FloatValue: 0.15,
+			Kind:       core.SpellMod_DamageDone_Pct,
+		})
+	}
```
rosenrusinov, line 155: Looking at the DB Shadowform comes up as a generic `Apply Aura: Mod Damage Done % (Shadow)` which should apply to ALL source of shadow damage and not just spells. So this should be `priest.PseudoStats.SchoolDamageDealtMultiplier[stats.SchoolIndexShadow] *= 1.15`

rosenrusinov: I'm still getting used to the spell mods but in reality what you've done will do the same but instead of applying it to the generic unit pseudo stat it would apply to every single shadow school spell the priest registers (even non class ones). I guess I just need to get used to this way of defining mods 😅

rosenrusinov (approved):
We can merge this as it is but in a follow up PR you should remove all the unneccesery castings of the ClassSpellMask now that its int64.

Also there are still a lot of places where sim.RandomFloat() is used instead of sim.Proc()

PR #49, opened 2024-04-06
[Core] Implement DoT haste scaling for Cataclysm

InDebt:
* Added a new APL to query the DoT frequency. This should allow for easy checks (Remaining Time < DoT Frequency) to allow re-applying DoTs without clipping them.
* All player DoTs should be affected by haste and thus automatically be affected by the patch
* The tick that gets carried over to the new dot also profits from the new snapshots - see sources in the commit

## Todo
* Further reasearch if all DoTs in general are affected by these changes or if there are exeptions and how HoTs behave

### DIFF
--- a/sim/core/dot.go
+++ b/sim/core/dot.go
@@ -144,6 +144,9 @@ func (dot *Dot) Apply(sim *Simulation) {
 		nextTick := dot.TimeUntilNextTick(sim)
 		dot.RecomputeAuraDuration()
 		dot.Aura.Duration += nextTick
+
+		// add extra tick
+		dot.TickCount--
 	} else {
 		dot.RecomputeAuraDuration()
 	}
--- a/sim/core/dot.go
+++ b/sim/core/dot.go
@@ -125,21 +125,29 @@ func (dot *Dot) RescheduleNextTick(sim *Simulation) {
 	sim.AddPendingAction(dot.tickAction)
 }
 
+// Snapshots and activates the Dot
+// If the Dot is running it's duration will be refreshed and
+// if there was a next Dot happening this will carry over to the new Dot
 func (dot *Dot) Apply(sim *Simulation) {
-
-	// for now all DoTs affected by haste (Player DoTs) should be affected
-	// by cata haste scaling no matter what
-	// TODO: Research how HoTs, physical DoTs and trinket induced DoTs might be handled
-	if dot.AffectedByCastSpeed {
-		dot.ApplyOrRefresh(sim)
-		return
-	}
 
 	dot.TakeSnapshot(sim, false)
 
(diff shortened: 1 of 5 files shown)


rosenrusinov:
Dots by melee classes are not hasted iirc (dk for example)

InDebt:
Yeah I should have been more specific.
Right now this change applies to all DoTs that have AffectedByHaste set to true. So this should not be the case for DK or physical DoTs.
Albeit small, the total duration of the DoT that gets rebuffed is extended by the next tick time. I'm just unsure if this now applies to Melee DoTs like deep wounds. In WoTLK it would just refresh the duration and keep the tick timer form the last DoT without the total duration being extended, right?

sim/core/dot.go
```diff
@@ -124,12 +125,29 @@ func (dot *Dot) RescheduleNextTick(sim *Simulation) {
...
 	dot.TakeSnapshot(sim, false)
 
-	dot.Cancel(sim)
 	dot.TickCount = 0
-	dot.RecomputeAuraDuration()
+
+	// we a have running dot tick
+	// the next tick never get's clipped and is added onto the dot's time for hasted dots
+	// see: https://github.com/wowsims/cata/issues/50
+	if dot.tickAction != nil {
```
rosenrusinov, line 140: `dot.tickAction != nil && !dot.tickAction.canceled`

Its safer to check if the action was canceled as well as not nil

sim/core/dot.go
```diff
@@ -124,12 +125,29 @@ func (dot *Dot) RescheduleNextTick(sim *Simulation) {
...
+
+		// save next tick timer as timer is computed based on tick time
+		// which we update in RecomputeAuraDuration
+		nextTick := dot.TimeUntilNextTick(sim)
+		dot.RecomputeAuraDuration()
+		dot.Aura.Duration += nextTick
```
rosenrusinov, line 146: Dont you need to also add an extra tick to the dot? Just increasing its duration will mean it wont tick the last time but just expire I think

InDebt: Yeah I confirmed it's indeed the case. My bad. 

I'd not like to just 'add' an extra tick however. Talents like Improved Devouring Plague require the correct amount of the DoT's ticks. So I see two options.

1. Add a new Option like `HasRollOverTick` which would add +1 in `MaxTicksRemaining`
2. Create a new utility method `ExpectedTickCount` on `Dot` which both `RecomputeAuraDuration` would call and spells / auras that need the dot's tick count. This would have the drawback however, that the tick count could be different if it's queried not on dot apply.

rosenrusinov: I wonder if you can just make the tickCount -1 😅

PR #51, opened 2024-04-08
[Core] Add SpellMods 

InDebt:
Lightweight SpellMod implmentation that should help implement proper separation of concerns for talents, buffs and spells.
The mod list might not be exhaustive and can be extended as needed.

### DIFF
--- a/sim/core/spell_mod.go
+++ b/sim/core/spell_mod.go
@@ -24,7 +24,6 @@ type SpellMod struct {
 	School         SpellSchool
 	floatValue     float64
 	intValue       int64
-	Stacks         int32
 	timeValue      time.Duration
 	Apply          SpellModApply
 	Remove         SpellModRemove
@@ -55,7 +54,6 @@ func buildMod(unit *Unit, config SpellModConfig) *SpellMod {
 		Apply:      functions.Apply,
 		Remove:     functions.Remove,
 		IsActive:   false,
-		Stacks:     1,
 	}
 
 	unit.OnSpellRegistered(func(spell *Spell) {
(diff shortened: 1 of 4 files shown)


sim/core/spell_mod.go
```diff
@@ -0,0 +1,272 @@
...
+func (unit *Unit) AddStaticMod(config SpellModConfig) {
+	mod := buildMod(unit, config)
+	mod.Activate()
+}
+
+// Never use dynamic mods for Auras that have ExpireNever and activate on reset
+// Those mods will be overwritten potentilly during sim reset
+func (unit *Unit) AddDynamicMod(config SpellModConfig) *SpellMod {
+	return buildMod(unit, config)
+}
```
rosenrusinov, line 81: Why is this a problem for NeverExpire and activation on reset?
Usually auras that are set to be permanent (aka neverexpire+activate on reset) are properly deactivated at the end of the iteration and activated at the start of the next.

rosenrusinov (approved):
One question about a comment but apart from that looks great

sim/core/spell_mod.go
InDebt: Yes. But before auras a disabled, we do Spell.Rest - and in spell.Reset certain parameters the aura modified are overwritten.

So it's basically like this right now.

Static Mods
Spell Snapshot (Initial Values)
Dynamic Mods
Spell Reset (Overwrite values like DamageMultiplier, DamageMultiplerAdd)
Aura Reset (Where we would remove mods we applied but those mods effects was removed in Spell Reset potentially)

PR #54, opened 2024-04-10
Update enchants and meta gems

InDebt:
## Successor to Neteyes PR #40

* Adds cataclysm meta gems and implementation
* Adds cataclysm enchants and implementation
* Some improvements for UI (OH enchants and item icons)
* Adds two new python scripts to scrape enchants and enchant descriptions


rosenrusinov (approved):
I havent looked in detail of the enchant/gem implementations but we will iterate on that when we have concrete game testing done

PR #57, opened 2024-04-11
[Races] Update cata sim for new race changes

InDebt:
* Adds worgen and their racial
* Adds goblin and their racial
* Update base stats for all races
* Update old racials to new cata version
* Add new race / class combinations


sim/core/racials.go
```diff
@@ -130,8 +130,18 @@ func applyRaceEffects(agent Agent) {
...
-		spBonus := float64(character.Level)*2 + 3
+		apBonus := 0.0
+		spBonus := 0.0
+
+		switch character.Class {
+		case proto.Class_ClassWarlock:
+			spBonus = 584.0
+		case proto.Class_ClassShaman:
+			spBonus = 584.0
+			spBonus = 1196.0
```
rosenrusinov, line 141: apBonus?

InDebt: yeah.. whoops

PR #58, opened 2024-04-11
[Shaman] Cata Updates

ghronkrepps:
I merged the changes into a new branch so I think I have to create a new pull request.
Sorry not an expert on Git. Copying relevant information here.

This is in a pretty good state. All Ele and Enhance talents, abilities, glyphs, etc are updated. Pet scaling will need worked on as we get better numbers.


sim/core/spell_mod.go
```diff
@@ -324,6 +324,14 @@ func removeCastTimePercent(mod *SpellMod, spell *Spell) {
...
 }
 
+func applyDotNumberOfTicksFlat(mod *SpellMod, spell *Spell) {
+	spell.CurDot().NumberOfTicks += int32(mod.intValue)
+}
+
+func removeDotNumberOfTicksFlat(mod *SpellMod, spell *Spell) {
+	spell.CurDot().NumberOfTicks -= int32(mod.intValue)
+}
+
```
rosenrusinov, line 334: This should be removed

proto/common.proto
```diff
@@ -431,25 +431,25 @@ message RaidBuffs {
 
 	// 10% AP
 	bool trueshot_aura = 11;
-	bool unleashed_rage = 12;
-	bool abominations_might = 13;
+	TristateEffect unleashed_rage = 12;
+	TristateEffect abominations_might = 13;
```
rosenrusinov, line 435: we use tristate effects for buffs that can be improved with a talent and not for points leveled. We always assume when external those talents are at max points. I think we want this to be back to bool and assume max bonus

sim/core/mana.go
```diff
@@ -33,11 +33,11 @@ type manaBar struct {
 // as well as enable the mana gain action to regenerate mana.
 // It will then enable mana gain metrics for reporting.
 func (character *Character) EnableManaBar() {
-	character.EnableManaBarWithModifier(1.0)
+	character.EnableManaBarWithModifier(1.0, true)
 	character.Unit.SetCurrentPowerBar(ManaBar)
 }
 
-func (character *Character) EnableManaBarWithModifier(modifier float64) {
+func (character *Character) EnableManaBarWithModifier(modifier float64, allowSpellPowerScaling bool) {
```
rosenrusinov, line 40: This can be reverted as we discussed in the discord cleaner ways to overwrite the spell power scaling

sim/core/spell_mod.go
ghronkrepps: got it

sim/core/mana.go
ghronkrepps: ok

proto/common.proto
ghronkrepps: so that would mean you could never sim if the dk in your raid only has one point in it?

PR #63, opened 2024-04-13
Add P1 Trinkets

InDebt:
Should include most of the P1 Trinkets.
Not included are:

* [Elementium Dragonling](https://www.wowhead.com/cata/item=60403/elementium-dragonling)
* [Corrupted Egg Shell](https://www.wowhead.com/cata/item=56136/corrupted-egg-shell)
* [Baradin Footman's Tags](https://www.wowhead.com/cata/item=65175/baradin-footmans-tags)

Updates some overwrite to exclude some trinkets that are not available ingame.


rosenrusinov (approved):
Love the new utility additions <3

InDebt:
Haha, most of it is just refactored from what was there for Wrath + some extension for dmg procs + custom conditions for a few trinkets.

rosenrusinov:
Only thing I see missing is for the damage proc helper to accept an extra parameter for the proc spell id. I know just casting the spell with the item id will work but we've found that users are actually confused when they see different spells in their logs and when trying to compare with the sim.

PR #65, opened 2024-04-14
[Mage] Cata Mage Baseline Spells 

gjoeyjoe:
- Most spells added.
- Living Bomb cap WIP
- Talents still WIP
- Glyphs still WIP

To do:
- [x] Combustion
- [x] Ignite
- [x] Water Elemental
- [x] Living Bomb cap on 3 targets
- [x] Talents and Glyphs just need to plugged away on
- [x] Mastery for Arcane and Frost
- [x] Flame Orb and Frostfire Orb need some spell IDs to get some numbers
- [x] Currently Flame Orb and Frostfire Orb are implemented as splitting each tick by number of enemies, makes it look like avg hit is lower than it really is, but overall damage should be correct. Ideally, would be set to pick a random target from available enemies, but I'm still new to all this and am not 100% sure on how best to implement that.


sim/mage/_ignite.go
```diff
@@ -67,34 +64,26 @@ func (mage *Mage) applyIgnite() {
...
+	switch mage.Talents.Ignite {
+	case 1:
+		igniteDamageMultiplier = 0.13
+	case 2:
+		igniteDamageMultiplier = 0.26
+	case 3:
+		igniteDamageMultiplier = 0.40
+	default:
+		return
+	}
```
rosenrusinov, line 79: `igniteDamageMultiplier := []float64{0.0, 0.26, 0.40}[mage.Talents.Ignite]`

rosenrusinov (commented):
You should port your spell mod implementations to the new SpellMod system we added for cata. This separates talents/glyphs/item mods out of spell implementations and into their own things which really helps the codebase. You can find examples of the new system in priest, warrior and dk code. We are slowly porting everything to use the new system

sim/mage/talents.go
```diff
@@ -122,15 +122,29 @@ func (mage *Mage) ApplyTalents() {
...
+			ClassMask:  MageSpellFireDoT,
+			FloatValue: float64(1.22 + 0.28*mage.GetMasteryPoints()),
+			Kind:       core.SpellMod_DamageDone_Pct,
+		})
+		fireMastery.Activate()
+
+		mage.AddOnMasteryStatChanged(func(sim *core.Simulation, oldMastery, newMastery float64) {
+			fireMastery.UpdateFloatValue(1.22 + 0.28*core.MasteryRatingToMasteryPoints(newMastery))
+		})
+	}
```
rosenrusinov: This is leaking spec specific code in the class package. Instead you should move this to the fire specs go files. Check DK for example setup

rosenrusinov: Also update should be `0.22 + ...` insted of `1.22` (The spell mod already adds 1)

sim/mage/fire/fire.go
```diff
@@ -31,6 +31,8 @@ func NewFireMage(character *core.Character, options *proto.Player) *FireMage {
 	}
 	fireMage.FireOptions = fireOptions
 
+	fireMage.ApplyTalents()
```
rosenrusinov: You dont need to call this here. Its automatically called by character

sim/mage/fire/fire.go
```diff
@@ -45,3 +47,45 @@ func (fireMage *FireMage) GetMage() *mage.Mage {
...
+
+	fireMage.registerPyroblastSpell()
+}
+
+func (fireMage *FireMage) GetMasteryBonus() float64 {
+	return (0.22 + 0.028*fireMage.GetMasteryPoints())
+}
+
+func (fireMage *FireMage) ApplyTalents() {
+
```
rosenrusinov, line 61: You need to call `fireMage.mage.ApplyTalents()` here

gjoeyjoe:
The current docket is:
**Major**
- [ ] Get pyromaniac functioning
- [ ] Impact spreading dots correctly. Right now it recasts them with the same damage. It should also recast them with the remaining duration and ticks. 

**Minor**
- [ ] Double check hot streak proc calculations against the `-2.7*critchance + 0.85` from EJ (https://web.archive.org/web/20120510120826/http://elitistjerks.com/f75/t104767-fire_cataclysm_discussion_op_updated_release/p15/#post1769730). A larger sample size would be beneficial to accurately derive the formula. 
- [ ] Double check water elemental disobeying. Was removed in PR since in Beta it seemed to chain cast fine. If it affects Freeze, may look at having something to reflect it, such as a randomized delay between Freeze cast and Freeze effect due to Elemental behavior.

sim/mage/arcane/arcane.go
```diff
@@ -31,13 +37,9 @@ func NewArcaneMage(character *core.Character, options *proto.Player) *ArcaneMage
 	}
 	arcaneMage.ArcaneOptions = arcaneOptions
 
-	return arcaneMage
-}
-
-type ArcaneMage struct {
-	*mage.Mage
+	arcaneMage.ApplyArcaneSpecInnate()
```
rosenrusinov, line 40: This should not be called from the NewArcaneMage func but from the initialize instead

sim/mage/arcane/arcane_barrage.go
```diff
@@ -0,0 +1,49 @@
...
+				GCD: core.GCDDefault,
+			},
+			CD: core.Cooldown{
+				Timer:    Mage.NewTimer(),
+				Duration: time.Second * 4,
+			},
+		},
+
+		DamageMultiplier: 1 +
+			.02*float64(Mage.Talents.TormentTheWeak),
```
rosenrusinov, line 35: move to spell mod

sim/mage/arcane_blast.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/wowsims/cata/sim/core/proto"
+)
+
+func (mage *Mage) registerArcaneBlastSpell() {
+
+	abDamageScalar := core.TernaryFloat64(mage.HasPrimeGlyph(proto.MagePrimeGlyph_GlyphOfArcaneBlast), 0.13, 0.1)
+	abDamageMod := mage.AddDynamicMod(core.SpellModConfig{
+		ClassMask:  MageSpellArcaneBlast | MageSpellArcaneExplosion,
+		FloatValue: abDamageScalar,
+		Kind:       core.SpellMod_DamageDone_Pct,
```
rosenrusinov, line 16: prob flat and not pct

sim/mage/arcane_blast.go
```diff
@@ -0,0 +1,79 @@
...
+	})
+	abCostMod := mage.AddDynamicMod(core.SpellModConfig{
+		ClassMask:  MageSpellArcaneBlast | MageSpellArcaneExplosion,
+		FloatValue: 1.5,
+		Kind:       core.SpellMod_PowerCost_Pct,
```
rosenrusinov, line 21: prob flat and not pct

sim/mage/blast_wave.go
```diff
@@ -28,18 +29,22 @@ func (mage *Mage) registerBlastWaveSpell() {
 				Duration: time.Second * 30,
 			},
 		},
-		BonusCritRating: float64(mage.Talents.CriticalMass+mage.Talents.WorldInFlames) * 2 * core.CritRatingPerCritChance,
 		DamageMultiplierAdditive: 1 +
-			.02*float64(mage.Talents.SpellImpact) +
-			.02*float64(mage.Talents.FirePower),
-		CritMultiplier:   mage.SpellCritMultiplier(1, mage.bonusCritDamage),
-		ThreatMultiplier: 1 - 0.1*float64(mage.Talents.BurningSoul),
+			.01*float64(mage.Talents.FirePower),
```
rosenrusinov, line 33: move to spell mod

sim/mage/evocation.go
```diff
@@ -8,21 +8,22 @@ import (
...
+		ClassSpellMask: MageSpellEvocation,
 
 		Cast: core.CastConfig{
 			DefaultCast: core.Cast{
 				GCD: core.GCDDefault,
 			},
 			CD: core.Cooldown{
 				Timer:    mage.NewTimer(),
-				Duration: time.Minute * time.Duration(4-mage.Talents.ArcaneFlows),
+				Duration: time.Minute*4 - time.Duration(mage.Talents.ArcaneFlows),
```
rosenrusinov, line 26: spell mod

sim/mage/fire/fire.go
```diff
@@ -31,6 +31,7 @@ func NewFireMage(character *core.Character, options *proto.Player) *FireMage {
 	}
 	fireMage.FireOptions = fireOptions
 
+	fireMage.ApplyFireSpecInnate()
```
rosenrusinov, line 34: this needs to be called in initialize not in the struct creation

rosenrusinov (commented):
Instead of calling these `ApplySpecInnate()` methods manually in the spec files you need to just rename them to ApplyTalents and inside of them call the `spec.Mage.ApplyTalents()`

sim/mage/arcane/arcane.go
```diff
@@ -80,12 +79,28 @@ func (arcaneMage *ArcaneMage) ApplyArcaneSpecInnate() {
...
 		OnExpire: func(aura *core.Aura, sim *core.Simulation) {
 			arcaneMastery.Deactivate()
 		},
 	}))
 
+	core.MakeProcTriggerAura(&arcaneMage.Unit, core.ProcTrigger{
+		Name:     "Arcane Mastery Mana Updater",
+		Callback: core.CallbackOnCastComplete,
+		Handler: func(sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
+			arcaneMastery.UpdateFloatValue(arcaneMastery.GetFloatValue())
```
rosenrusinov, line 94: I think you copy/pasted the wrong func here and want to use `arcaneMage.ArcaneMasteryValue()`

gjoeyjoe: oops 🙃fixed

gjoeyjoe:
Have a potential contributor who wants to help, so hoping to have PR merged and they can fork from here.

sim/mage/arcane/arcane.go
```diff
@@ -34,16 +40,67 @@ func NewArcaneMage(character *core.Character, options *proto.Player) *ArcaneMage
...
+
+func (arcaneMage *ArcaneMage) ApplyTalents() {
+
+	arcaneMage.Mage.ApplyTalents()
+	// Arcane Specialization Bonus
+	arcaneMage.AddStaticMod(core.SpellModConfig{
+		School:     core.SpellSchoolArcane,
+		FloatValue: 0.25,
+		Kind:       core.SpellMod_DamageDone_Flat,
+	})
```
rosenrusinov, line 65: This looks weird as a mod. Usually Additive mods target specific spells. I'm not sure how this specialization should work but if its supposed to work with non mage spells like trinkets/procs it should probably use `DamageDone_Pct`. If it is indeed supposed to only affect mage spells then Flat + a class mask for MageSpellsAll would be correct

rosenrusinov: Looking at the spell in wowhead it does look like it hsould be `DamageDone_Pct` on all arcane spells not only mage ones

sim/mage/arcane/arcane.go
```diff
@@ -34,16 +40,67 @@ func NewArcaneMage(character *core.Character, options *proto.Player) *ArcaneMage
...
+		Callback: core.CallbackOnCastComplete,
+		Handler: func(sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
+			arcaneMastery.UpdateFloatValue(arcaneMage.ArcaneMasteryValue())
+		},
+	})
+
+}
+
+func (arcaneMage *ArcaneMage) GetArcaneMasteryBonus() float64 {
+	return (1.12 + 0.015*arcaneMage.GetMasteryPoints())
```
rosenrusinov, line 101: `0.12 +`

sim/mage/arcane_explosion.go
```diff
@@ -0,0 +1,38 @@
...
+		CritMultiplier:   mage.DefaultSpellCritMultiplier(),
+		BonusCoefficient: 0.186,
+		ThreatMultiplier: 1 - 0.4*float64(mage.Talents.ImprovedArcaneExplosion),
+
+		ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
+			for _, aoeTarget := range sim.Encounter.TargetUnits {
+				baseDamage := 0.368 * mage.ScalingBaseDamage
+				baseDamage *= sim.Encounter.AOECapMultiplier()
+				spell.CalcAndDealDamage(sim, aoeTarget, baseDamage, spell.OutcomeMagicHitAndCrit)
+			}
```
rosenrusinov, line 35: Aoe spells like this are usually first all results calculated and then all results dealt in a 2nd loop. You can find many examples in other classes

sim/mage/arcane_missiles.go
```diff
@@ -0,0 +1,86 @@
...
+					if dot.TickCount < dot.NumberOfTicks {
+						dot.TickCount++
+						dot.TickOnce(sim)
+					}
+					mage.ArcaneBlastAura.Deactivate(sim)
+				},
+			},
+			NumberOfTicks:       numTicks - 1, // subtracting 1 due to autocasting one OnGain
+			TickLength:          time.Millisecond * 700,
+			AffectedByCastSpeed: true,
```
rosenrusinov, line 73: I'm not sure if arcane missiles work like other hasted dots. Is this supposed to keep channeling for the original duration and just send more missiles if you have more haste? Thats whats gonna happen with the current dot system.

sim/mage/combustion.go
```diff
@@ -0,0 +1,139 @@
...
+						if dots != nil && dots.IsActive() {
+							normalizedDPS = 1000000000 * spell.Dot(mage.CurrentTarget).SnapshotBaseDamage / float64(spell.Dot(mage.CurrentTarget).TickPeriod())
+
+						}
+					}
+					combustionDotDamage += normalizedDPS
+				} */
+				var spellDPS float64
+				for _, spell := range dotSpells {
+					dots := spell.Dot(mage.CurrentTarget)
```
rosenrusinov, line 62: you should use the `target` from the arguments instead of CurrentTarget

sim/mage/combustion.go
```diff
@@ -0,0 +1,139 @@
...
+		},
+		DamageMultiplierAdditive: 1,
+		CritMultiplier:           mage.DefaultSpellCritMultiplier(),
+		BonusCoefficient:         1.113,
+		ThreatMultiplier:         1,
+
+		ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
+			baseDamage := 0.429 * mage.ScalingBaseDamage
+			result := spell.CalcDamage(sim, target, baseDamage, spell.OutcomeMagicHit)
+			spell.DealDamage(sim, result)
```
rosenrusinov, line 93: I think this should be in the Impact spell

sim/mage/combustion.go
```diff
@@ -0,0 +1,139 @@
...
+			OnTick: func(sim *core.Simulation, target *core.Unit, dot *core.Dot) {
+				dot.CalcAndDealPeriodicSnapshotDamage(sim, target, dot.OutcomeSnapshotCrit)
+			},
+		},
+		DamageMultiplierAdditive: 1,
+		CritMultiplier:           mage.DefaultSpellCritMultiplier(),
+		ThreatMultiplier:         1,
+
+		ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
+			spell.Dot(target).ApplyOrReset(sim)
```
rosenrusinov, line 135: If combustion can miss you should check here after you move the impact damage. (It looks like it has a hit check from your code)

sim/mage/evocation.go
```diff
@@ -8,21 +8,29 @@ import (
...
 	manaMetrics := mage.NewManaMetrics(actionID)
 	manaPerTick := 0.0
 
+	if mage.Talents.ArcaneFlows > 0 {
+		mage.AddStaticMod(core.SpellModConfig{
+			ClassMask: MageSpellEvocation,
+			TimeValue: -time.Minute * time.Duration(mage.Talents.ArcaneFlows),
+			Kind:      core.SpellMod_Cooldown_Flat,
+		})
+	}
```
rosenrusinov, line 21: This should be moved to talents files

sim/mage/fire/pyroblast.go
```diff
@@ -0,0 +1,146 @@
...
+			OnTick: func(sim *core.Simulation, target *core.Unit, dot *core.Dot) {
+				dot.CalcAndDealPeriodicSnapshotDamage(sim, target, dot.OutcomeTick)
+			},
+		},
+
+		ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
+			spell.Dot(target).ApplyOrReset(sim)
+			Mage.PyroblastDot.SpellMetrics[target.UnitIndex].Casts = 0
+		},
+	})
```
rosenrusinov, line 145: Why do you need a fake spell?

sim/mage/fire_blast.go
```diff
@@ -0,0 +1,44 @@
...
+				GCD: core.GCDDefault,
+			},
+			CD: core.Cooldown{ // Note: Impact talent triggers CD refresh on spell *land*, not cast
+				Timer:    mage.NewTimer(),
+				Duration: time.Second * 8,
+			},
+		},
+
+		BonusCritRating: 0 +
+			float64(mage.Talents.ImprovedFireBlast)*4*core.CritRatingPerCritChance,
```
rosenrusinov, line 31: spell mod in talents

sim/mage/flame_orb.go
```diff
@@ -0,0 +1,160 @@
...
+		CritMultiplier:   mage.DefaultSpellCritMultiplier(),
+		BonusCoefficient: 0.193,
+		ThreatMultiplier: 1,
+		ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
+
+			damage := 1.318 * mage.ScalingBaseDamage
+
+			for _, aoeTarget := range sim.Encounter.TargetUnits {
+				spell.CalcAndDealDamage(sim, aoeTarget, damage, spell.OutcomeMagicHitAndCrit)
+				spell.SpellMetrics[target.UnitIndex].Hits++
```
rosenrusinov, line 68: Why are you manually increasing hit metrics here?

sim/mage/flame_orb.go
```diff
@@ -0,0 +1,160 @@
...
+				GCD: core.GCDDefault,
+			},
+			CD: core.Cooldown{
+				Timer:    mage.NewTimer(),
+				Duration: time.Minute,
+			},
+		},
+
+		ApplyEffects: func(sim *core.Simulation, _ *core.Unit, _ *core.Spell) {
+			mage.flameOrb.EnableWithTimeout(sim, mage.flameOrb, time.Second*15)
```
rosenrusinov, line 39: I'm not sure if this spell requires to be modeled as a pet. If you really want to be able to show damage from another source unit then sure but it adds some complexities. Right now you're not getting any damage bonuses that the mage would have from temporary buffs and effects for example. Is this expected?

sim/mage/flamestrike.go
```diff
@@ -0,0 +1,62 @@
...
+		Flags:          SpellFlagMage | core.SpellFlagAPL,
+		ClassSpellMask: MageSpellFlamestrike,
+
+		ManaCost: core.ManaCostOptions{
+			BaseCost: 0.30,
+		},
+		Cast: core.CastConfig{
+			DefaultCast: core.Cast{
+				GCD:      core.GCDDefault,
+				CastTime: time.Millisecond*2000 - 1000*time.Duration(mage.Talents.ImprovedFlamestrike),
```
rosenrusinov, line 24: spell mod in talents

sim/mage/freeze.go
```diff
@@ -0,0 +1,47 @@
...
+			for _, aoeTarget := range sim.Encounter.TargetUnits {
+				baseDamage := 0.409 * mage.ScalingBaseDamage
+				baseDamage *= sim.Encounter.AOECapMultiplier()
+				spell.CalcAndDealDamage(sim, aoeTarget, baseDamage, spell.OutcomeMagicHitAndCrit)
+			}
+			if mage.FingersOfFrostAura.IsActive() {
+				mage.FingersOfFrostAura.Refresh(sim)
+			} else {
+				mage.FingersOfFrostAura.Activate(sim)
+			}
```
rosenrusinov, line 43: just call `mage.FingersOfFrostAura.Activate(sim)`. It will just refresh if its already active

sim/mage/frost/frost.go
```diff
@@ -23,25 +23,90 @@ func RegisterFrostMage() {
...
+	waterElemental *WaterElemental
+}
+
 func NewFrostMage(character *core.Character, options *proto.Player) *FrostMage {
 	frostOptions := options.GetFrostMage().Options
 
 	frostMage := &FrostMage{
-		Mage: mage.NewMage(character, options, frostOptions.ClassOptions),
+		Mage:             mage.NewMage(character, options, frostOptions.ClassOptions),
+		ClassBaseScaling: 937.330078125,
```
rosenrusinov, line 38: use the util function for getting the base scaling `core.GetClassSpellScalingCoefficient`

sim/mage/frost/frost.go
```diff
@@ -23,25 +23,90 @@ func RegisterFrostMage() {
...
+	frostMage.waterElemental.AddStaticMod(core.SpellModConfig{
+		School:     core.SpellSchoolFrost,
+		FloatValue: 0.25,
+		Kind:       core.SpellMod_DamageDone_Pct,
+	})
+
+	frostMage.Mage.AddStaticMod(core.SpellModConfig{
+		ClassMask:  mage.MageSpellFrostbolt,
+		FloatValue: 0.15,
+		Kind:       core.SpellMod_DamageDone_Pct,
```
rosenrusinov, line 78: prob `_Flat` instead of `_Pct`

sim/mage/arcane/arcane.go
gjoeyjoe: ah yeah. remnant from me testing if things were getting affected by it

sim/mage/arcane_missiles.go
gjoeyjoe: It should be a static number of ticks, standard 3, up to 5 based on talents. in testing this implementation, adding haste only shortens the duration and does not add ticks.

sim/mage/fire/pyroblast.go
gjoeyjoe: This was a testing thing for the Impact talent. My thought process was to have an Impact version of the spells that could be filled out each time it needed to be spread since it had custom duration. Ended up scrapping that idea half way through and am looking to find a way to duplicate existing auras onto the applicable targets instead, but haven't had time.

sim/mage/combustion.go
gjoeyjoe: Just asked and mage discord says combustion can miss

sim/mage/combustion.go
gjoeyjoe: Impact portion is from the previously mentioned Impact talent testing. Probably not a bad idea to split up combustion into the application damage and the dot ticks though, so may repurpose it.

sim/mage/flame_orb.go
gjoeyjoe: The pet version was initially written as such because it executes a rotation where every 1s for 15s it picks a (in practice) random target and hits them. It didn't make the most intuitive sense as a dot since it a) wouldn't make sense to show it as an aura on the enemy and b) would need to be able to randomly do damage to nearby targets. It should be remodeled to be able to benefit from buffs once there's time to do so.

sim/mage/arcane_missiles.go
rosenrusinov: Interesting... I would've expect it to work like any other dot we have which would add more ticks to fit the duration

sim/mage/flame_orb.go
gjoeyjoe: If there's a way to hide auras from showing up (i'm sure there is), then there's probably a good way to have an unhasted dot on an enemy that does the spellcast using the random-target-selection I'm currently using.

sim/mage/arcane_missiles.go
gjoeyjoe: Actually you're right, it just shows up differently than I expected in the timeline. Logs do show rapid casting, but timeline only shows 5 "ticks" causing a cast.

sim/mage/flame_orb.go
gjoeyjoe: idk, that was dumb.

sim/mage/flame_orb.go
rosenrusinov: I think you should make it an aura on the mage with an attached PeriodicAction that does your random target spell hit. This would make it nicer on the timeline as well as you can see when you have your flame orb active with other player auras

rosenrusinov: I can take a look at this once this PR is merged and clean it up a bit

sim/mage/frost/frost.go
```diff
@@ -23,25 +23,90 @@ func RegisterFrostMage() {
 	)
 }
 
+type FrostMage struct {
+	*mage.Mage
+	ClassBaseScaling float64
```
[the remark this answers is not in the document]
gjoeyjoe: This was due to a misunderstanding of how the pet would inherit the scaling stat. Fixed

PR #88, opened 2024-04-16
[Shadow] Alpha Release

InDebt:
This PR should bring Shadow into an Alpha stage where I feel confident that we can get valuable feedback based on the sims performance from anyone using it. All the foundational staff that would impact the sim in a meaningful way should be present.

## Defaults added for this PR
* Raid BiS Set
* Talents
* Consumes / Buffs and so on
* APL

## Fixes
* Priest Tier Sets (T11 included)
* Threat multipliers
* Introduction of armor specialization
* Variance on MB
* Implement Mind Spike
* Devouring Plague instant damage haste calculations
* Twisting Faith no longer gets applied twice

## Misc Fixes
* Remove level based resistances
* Fixes the stat selection for Ingi Hands
* Fixed Offhand Enchant


sim/priest/items.go
```diff
@@ -138,10 +175,15 @@ var ItemSetCrimsonAcolyte = core.NewItemSet(core.ItemSet{
...
+			character := agent.GetCharacter()
+			character.AddStaticMod(core.SpellModConfig{
+				Kind:       core.SpellMod_BonusCrit_Rating,
+				FloatValue: 5 * core.CritRatingPerCritChance,
+				ClassMask:  PriestSpellShadowWordPain | PriestSpellDevouringPlague | PriestSpellVampiricTouch | PriestSpellImprovedDevouringPlague,
+			})
 		},
 		4: func(agent core.Agent) {
-			// Implemented in mind_flay.go
+			// TODO: Implemement proper aura mod to mod total duration
```
rosenrusinov, line 186: Spells have a field called `RelatedAuras` which you can assign aura references to be kept in the spell so we can add a spell mod for aura duration and modify the duration of all related auras of that spell

InDebt: Honestly I forgot about this place holder. The correct wording would be 'Shorten tick period' i guess. As each tick is shortened by 0.17 and we got 3 ticks so a total of 0.5 seconds reduction. Wasn't sure how to properly implement that but I could do the same as you did for the SpelMod_NumTicks i guess.

InDebt: Done.

rosenrusinov (approved):

PR #97, opened 2024-04-16
[Hunter] Hunter Updates

ToxicKevinFerm:
Gonna continue work on SpellMods in another branch to keep them separate, it'll be quite significant changes it looks like.

This is more fixing some abilities, working on pet and updating apl&gear sets


rosenrusinov (approved):
I think I saw some glyphs with damage buffs to spells that are added with DamageDone_Pct instead of DamageDone_Flat but i'm no hunter so maybe its correct

sim/hunter/glyphs.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"github.com/wowsims/cata/sim/core"
+	"github.com/wowsims/cata/sim/core/proto"
+)
+
+func (hunter *Hunter) ApplyGlyphs() {
+	if hunter.HasPrimeGlyph(proto.HunterPrimeGlyph_GlyphOfArcaneShot) {
+		hunter.AddStaticMod(core.SpellModConfig{
+			Kind:       core.SpellMod_DamageDone_Pct,
```
InDebt, line 11: https://wago.tools/db2/SpellEffect?build=4.4.0.54217&filter[SpellID]=%0956841&page=1 as Rain pointed out, this should be DamageDone_Flat

ToxicKevinFerm: Ah, I definitely thought that `SpellMod_DamageDone_Pct` was the additive mod, will fix it.

PR #102, opened 2024-04-17
Add support for Dark Intent

InDebt:
* Make Dark Intent available as Personal Buff in the UI
* Add option to simulate different Dmg% uptimes, espacially for warlock sims later
* Add DotDamageDone% Pseudostat


ui/core/components/other_inputs.ts
```diff
@@ -101,6 +101,17 @@ export const ChannelClipDelay = {
 	},
 };
 
+export const DarkIntentUptime = {
+	type: 'number' as const,
+	label: 'Dark Intent Uptime',
+	labelTooltip: '% uptime on Dark Intent on the player (Only the stacking damage component)',
+	changedEvent: (player: Player<any>) => player.miscOptionsChangeEmitter,
```
rosenrusinov, line 108: You should add a showWhen property and only show this input when dark intent is enabled. Also if you do that you will need to update the changedEvent to also return the buff change emitter ontop of the misc options

PR #104, opened 2024-04-17
Fixup int / crit scaling for cata

InDebt:


sim/core/mana.go
```diff
@@ -48,6 +48,12 @@ func (character *Character) EnableManaBarWithModifier(modifier float64) {
 	// Starting with cataclysm 1 intellect now provides 1 spell power
 	character.AddStatDependency(stats.Intellect, stats.SpellPower, 1.0)
 
+	if character.Unit.Type == PlayerUnit {
+		// Every caster gains 1% crit per 648.91
+		// Pets have different scaling so let them handle their scaling
+		character.AddStatDependency(stats.Intellect, stats.SpellCrit, 1.0/648.91*CritRatingPerCritChance)
```
rosenrusinov, line 54: Do we maybe want to define this constant next to the other crit related constants? I know we have magic values sprinkled all around so just a suggestion at this point

rosenrusinov (approved):

PR #114, opened 2024-04-18
Duration of pots should be 25seconds

ToxicKevinFerm:


rosenrusinov (commented):
I dont think we want to remove flame cap at this point

ToxicKevinFerm:
> I dont think we want to remove flame cap at this point

Yeah thats not supposed to be there. adding it back..

Issue #94, opened 2024-04-16
Mana Exceeding 100%


s4lin0:
Description:
On Shadowpriest Cata Alpha Simulator mana can exceed 100%.

Reproduction:
Start the default sim.
Look at the mana changing under "Results" and "Timeline".
E. g. Shadowfiend or mana regenration tick raise current mana.
Current mana over 100%.

Expected behavior:
Mana is capped at 100% and should not exceed it.

Screenshot:
![manaCap](https://github.com/wowsims/cata/assets/167233741/3b6932c8-32c8-4f52-bfb9-5067f0f8e2d6)

PR #129, opened 2024-04-20
[Core] Fixup mana being able to go above 100%

InDebt:
Should address #94


kayla-glick (approved):
This would be a great change to port back to the SoD and WOTLK versions of the site if you wouldn't mind

ui/core/proto_utils/logs_parser.tsx
```diff
@@ -771,6 +776,14 @@ export class ResourceChangedLogGroup extends SimLog {
 	static fromLogs(logs: Array<SimLog>): Record<ResourceType, Array<ResourceChangedLogGroup>> {
 		const allResourceChangedLogs = logs.filter((log): log is ResourceChangedLog => log.isResourceChanged());
 
+		const maxRessource = function(logs: ResourceChangedLog[]) {
```
kayla-glick, line 779: 🔨 Small knit: Typo `Ressource` instead of `Resource`

ui/core/proto_utils/logs_parser.tsx
```diff
@@ -771,6 +776,14 @@ export class ResourceChangedLogGroup extends SimLog {
...
+			let max : number | null = null
```
kayla-glick, line 780: Could potentially allow for this to just be a `number` by adding totals to the other resource types too. Might clean it up a bit but I don't feel that strongly about it

PR #132, opened 2024-04-20
Warrior specs implementation

Swillis57:
Also includes some spell scaling coefficient helpers in utils.go as well as a new Multiplier field on core.Cooldown and a corresponding SpellMod. 

Warrior has effects that modify CDs by flat amounts as well as percentages (on the same ability, see Shield Wall's glyph and the T11 tank set). A previous implementation that multiplied CD.Duration directly was not safe to use alongside Cooldown_Flat as the final CD would be dependent on the order the mods were applied in ((5min*0.5) - 2min = 30sec,  (5min - 2min)*0.5 = 1.5min). The multiplier is currently accumulated additively to allow reaching 0 CD, though I don't believe it's possible to have multiple CD reduction multipliers at once in Cata. 

The spell scaling helpers are really just an effort to get them somewhere they can be found so we don't need to keep referencing random DB tables and discord posts for an already obscure mechanic. Their actual uses are waiting on the beta build actually getting the missing variance field filled in.


Swillis57:
The autoformat in VS code seems to have poked some files I didn't actually edit

sim/warrior/arms/arms.go
```diff
@@ -129,20 +129,10 @@ func (war *ArmsWarrior) GetWarrior() *warrior.Warrior {
...
 }
 
-// func (war *ArmsWarrior) Reset(sim *core.Simulation) {
-// 	war.Warrior.Reset(sim)
-// 	war.BattleStanceAura.Activate(sim)
-// 	war.Stance = warrior.BattleStance
-// }
+func (war *ArmsWarrior) Reset(sim *core.Simulation) {
+	war.Warrior.Reset(sim)
+	war.BattleStanceAura.Activate(sim)
```
rosenrusinov, line 137: Stance activation should be done from prepull actions instead of being hardcoded here.

Swillis57: Oh I meant to remove these actually, thanks for pointing it out

sim/warrior/arms/deadly_calm.go
```diff
@@ -0,0 +1,66 @@
...
+	if !war.Talents.DeadlyCalm {
+		return
+	}
+
+	dcActionID := core.ActionID{SpellID: 85730}
+
+	// Using an aura that modifies the PseudoStat instead of a SpellMod to avoid stomping
+	// over Battle Trance if it's procced. Having both of these edit spells could result in a sequence like
+	// 1: <Use Deadly Calm> (spell.CostMultiplier = 0)
+	// 2: <Proc Battle Trance> (spell.CostMultiplier = -1) (what does this even do? refund on every ability?)
```
InDebt, line 20:
```Go
func (spell *Spell) ApplyCostModifiers(cost float64) float64 {
	cost -= spell.Unit.PseudoStats.CostReduction
	cost = max(0, cost*spell.Unit.PseudoStats.CostMultiplier)
	return max(0, cost*spell.CostMultiplier)
}
```
That's what it would do. So -1 wouldn't be that bad

sim/warrior/arms/mortal_strike.go
```diff
@@ -0,0 +1,49 @@
...
+				GCD: core.GCDDefault,
+			},
+			CD: core.Cooldown{
+				Timer:    war.NewTimer(),
+				Duration: time.Millisecond * 4500,
+			},
+		},
+
+		DamageMultiplier: 1.0,
+		CritMultiplier:   war.DefaultMeleeCritMultiplier(),
```
InDebt, line 34: Not sure if anyone ever is going to use this, but you are not setting the `ThreatMultiplier`

sim/warrior/arms/talents.go
```diff
@@ -0,0 +1,200 @@
...
+	// Use a specific aura for TfB so we can track procs
+	// Overpower will check for any aura with the EnableOverpowerTag when it tries to cast
+	tfbAura := war.RegisterAura(core.Aura{
+		Label:    "Taste for Blood",
+		ActionID: core.ActionID{SpellID: 60503},
+		Duration: time.Second * 9,
+		Tag:      warrior.EnableOverpowerTag,
+	})
+
+	core.MakePermanent(war.RegisterAura(core.Aura{
```
InDebt, line 51: You could use one of the core Helpers here

```Go
		core.MakeProcTriggerAura(war, core.ProcTrigger{
			Name: "Taste for Blood Trigger",
			Callback: core.CallbackOnPeriodicDamageDealt,
			ClassSpellMask: war.Rend,
			ICD: time.Second * 5,
			ProcChance: procChance,
			Handler: func(sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
				tfbAura.Activate(sim)
			},
		})
		```

File: sim/warrior/arms/mortal_strike.go (hunk 3 above)
Comment (Swillis57) [reply sim/warrior/arms/mortal_strike.go:34]: Yeah, I'm not really sure what to set it to and I don't think we track threat for DPS sims

File: sim/warrior/arms/blood_frenzy.go [hunk 5]
```diff
@@ -25,31 +24,32 @@ func (war *ArmsWarrior) applyBloodFrenzy() {
...
-
-				// Apply Trauma, has fixed duration regardless of bleeds
-				trauma := traumaAuras.Get(result.Target)
-				trauma.Activate(sim)
-			}
+	core.MakeProcTriggerAura(&war.Unit, core.ProcTrigger{
+		Name:       "Blood Frenzy Rage Trigger",
+		Callback:   core.CallbackOnSpellHitDealt,
+		Outcome:    core.OutcomeLanded,
+		ProcMask:   core.ProcMaskMeleeMHAuto,
```
InDebt, line 31: Only MH attacks should trigger this? Or is this just because it being an arms talent? :D There's no way some one dual wields arms, right? :D

sim/warrior/talents.go
```diff
@@ -232,28 +224,27 @@ func (warrior *Warrior) applyBloodAndThunder() {
...
+	core.MakeProcTriggerAura(&warrior.Unit, core.ProcTrigger{
+		Name:           "Blood and Thunder Trigger",
+		Callback:       core.CallbackOnSpellHitDealt,
+		Outcome:        core.OutcomeLanded,
+		ClassSpellMask: SpellMaskThunderClap,
+		ProcChance:     procChance,
+		ExtraCondition: func(sim *core.Simulation, spell *core.Spell, result *core.SpellResult) bool {
+			// If the rend we're checking was applied this iteration, skip to avoid an explosion of B&T procs
+			// (8 targets, Rend T1, TClap hits T1 + B&T applies Rend to 7 other targets, TClap hits T2 + applies Rend to 7 other targets, etc...)
+			return result.Target.HasActiveAuraWithTag("Rend") && lastAppliedTime != int64(sim.CurrentTime)
```
InDebt, line 236: fyi Aura has `StartedAt()` baked in already, but should not matter much

PR #135, opened 2024-04-20
[Shaman] Fixing LavaBurst clipping and adding Enhance Presets

foreverpsycotic:
Adding stop attack logic from sod so auto attacks don't clip


InDebt:
There seem to be some change from syntax as well on the matter of gear and presets. You should discuss what should be adopted. Alot of things seem to be the same 🤔

ghronkrepps:
We discussed on discord. Basically he gave me his gearset/apl and I updated into my PR. He tweaked it a bit and then pushed the changes in a separate branch. Also one fix for enhance casting during spells is fixed here.

@foreverpsycotic Since the APL/gear changed make sure the tests are running with your changes then update tests and then commit those with this.

Issue #138, opened 2024-04-20
SpellMod_DotNumberOfTicks_Flat not working for hasted dots


ghronkrepps:
**Describe the bug**
Flame Shock is a hasted dot and is not benefiting from the glyph as it is implemented because the spell mod SpellMod_DotNumberOfTicks_Flat not working for hasted dots. Discussed a bit in the shaman cata channel.

**To Reproduce**
Run the Shaman sim with and without the glyph there is no difference in length of the dot.

**Expected behavior**
Dot should be gaining 3 ticks/9 seconds.

PR #139, opened 2024-04-20
[Core] Refactor DoT haste handling

InDebt:
* Should now properly allow tick mods to change hasted spells tick count


InDebt:
Fixes #138

PR #154, opened 2024-04-22
[Rogue] Bug fixes and add tests

TheBackstabi:
Title.

With tests implemented and bugs largely addressed, next is refactoring for SpellMods.


sim/core/attack.go
```diff
@@ -643,6 +643,26 @@ func (aa *AutoAttacks) DelayMeleeBy(sim *Simulation, delay time.Duration) {
 	}
 }
 
+// PauseMeleeBy will prevent any swing from completing for the specified time.
+// This replicates a /stopattack and /startattack with a brief "pause" in the middle.
+// It's possible that no swing time is lost if the pauseTime is less than the remaining swing time.
+// Used by Rogue Gouge
+func (aa *AutoAttacks) PauseMeleeBy(sim *Simulation, pauseTime time.Duration) {
```
InDebt, line 650: What's the difference to the existing `StopMeleeUntil` ?

sim/rogue/recuperate.go
```diff
@@ -36,25 +36,37 @@ func (rogue *Rogue) registerRecuperate() {
...
+				healValue := rogue.MaxHealth() * (.03 + .005*float64(rogue.Talents.ImprovedRecuperate))
+				dot.Spell.CalcAndDealPeriodicHealing(sim, target, healValue, dot.OutcomeTick)
 
 				if rogue.Talents.EnergeticRecovery > 0 {
 					energyRegen := float64(rogue.Talents.EnergeticRecovery) * 4.0
-					rogue.AddEnergy(sim, energyRegen, energeticRecoveryMetrics)
+					// Trigger Energetic Recovery after small delay to prevent aura refresh loops
+					// https://i.gyazo.com/dc845a371102294abfb207c6fd586bfa.png
+					core.StartDelayedAction(sim, core.DelayedActionOptions{
+						DoAt:     sim.CurrentTime + 1,
```
InDebt, line 56: Might not be needed anymore with the new energy changes and you can even schedule it for the same time. Although 1 nanosecond won't make the difference I guess x).

sim/core/attack.go
TheBackstabi: `StopMeleeUntil` always changes the time the weapon should swing at (this effectively stops the swing timer from counting down); `PauseMeleeBy` only updates the `swingAt` if it would've occurred during that pause.

I manually confirmed this is how Gouge should operate, so it needed this specific functionality.

TheBackstabi:
Should be good to go here, haven't identified any other bugs and should get tests merged in

PR #166, opened 2024-04-23
[Balance] Add support for Lunar/Solar Power Bar

InDebt:
## Proposal for Lunar/Solar Power Bar for Balance Druids

So this is a draft so ppl who actually know balance druids can give feedback on this concept before I merge it :D.

### How it works ingame

Balance druids ingame gain a secondary power bar. This bar starts out at neutral position. Using `Lunar Energy` or `Solar Energy` generators the bar can be moved towards one direction.
* Wrath will add $13\frac{1}{3}$ `Lunar Energy` [WoW Head Reference](https://www.wowhead.com/cata/spell=5176/wrath#comments:id=1337613)
* Starfire will add $20$ `Solar Energy`
* Star Surge will ad $15$ `Solar` or `Lunar Energy` depending in which eclipse cycle you are

You start in a neutral state where you can either generate `Solar` or `Lunar Energy`. If you gain 20 through Starfire, casting Wrath in this state will move the bar in the other direction, so you end up with 7 `Solar Energy`. Upon reaching 100 Energy the druid enters the eclipse state where he can no longer generate the same energy until they reach the other 100 energy point (so they need to generate 200 energy of the opposite type in total). Thus, they are forced to cycle between Lunar and Solar eclipse.

From 100 to 0 energy the druid gains the respective Eclipse's buff.

### Implementation

To rebuild this in the core, i've added a new `eclipse` bar. It tracks solar and lunar energy, keeps track of which energy can currently be earned and offers callbacks for reaching the various Eclipse state those are.

* Enter Lunar Eclipse (Reaching 100 Energy)
* Exit Lunar Eclipse (Reaching 0 Energy)
* Enter Solar Eclipse (Reaching 100 Energy)
* Exit Solar Eclipse (Reaching 0 Energy)

### APL Support

druid_eclipse_support

I have added three APL functions to support proper APL's for balance druids.
* Query current Solar Energy
* Query current Lunar Energy
* Act based on current Eclipse state
  * Contrary to the Eclipse Buff, the Eclipse Phase describes the whole phase in which no other energy can be gained. So from 100 Lunar Energy to 100 Solar energy and the other way around.

I have played a bit of visualizing the current eclipse bar similar to ingame.
grafik


rosenrusinov (approved):
Can't say anything about the mechanics themself but the code looks good

InDebt:
Well everyone else had their fair share of time to critique. So thill will be V1 and then druid dev can update stuff that needs to be changed I guess.

PR #178, opened 2024-04-24
[Shaman] Fire elemental number update and 1 crash fix

Polynomix:


raethkcj:
To fix the failing test, you can locally run `make test`, confirm the only failing tests are ones you expected to change, and then `make update-tests`. Then add, commit, and push the changed test file, and the action here should run again and pass

PR #185, opened 2024-04-24
Hunter fix 24/04

ToxicKevinFerm:
1. Adds `OnFocusGain` callback to `focusBar`
2. Uses owners RAP for base damage on pet focus dump
3. Add Aspect of the Fox for simulating movement dps
4. Fix Wild Hunt, it needs to know focus values before spell cast, uses `OnFocusGain`
5. Add base crit scaling for pets in accordance with beta
6. Disable 3% damage raid buff on pets (i dont think it should be active, technically I dont think 10% haste should be either but it matches simc values
7. Fixes Sic'Em talent


rosenrusinov (approved):
We tested on beta the 10% and it actually applies to pets and double dips with the attack speed they get from the owner getting the buff as well

PR #197, opened 2024-04-25
[Warrior] Added Heroic leap

MatthewKyle:
Added Heroic Leap

### DIFF
--- a/sim/warrior/heroic_leap.go
+++ b/sim/warrior/heroic_leap.go
@@ -1,0 +1,50 @@
+package warrior
+
+import (
+	"time"
+
+	"github.com/wowsims/cata/sim/core"
+)
+
+func (warrior *Warrior) RegisterHeroicLeap() {
+
+	numHits := warrior.Env.GetNumTargets()
+	warrior.RegisterSpell(core.SpellConfig{
+		ActionID:       core.ActionID{SpellID: 6544},
+		SpellSchool:    core.SpellSchoolPhysical,
+		ProcMask:       core.ProcMaskMeleeMHSpecial,
+		Flags:          core.SpellFlagMeleeMetrics | core.SpellFlagAPL,
+		ClassSpellMask: SpellMaskHeroicLeap | SpellMaskSpecialAttack,
+
+		Cast: core.CastConfig{
+			DefaultCast: core.Cast{},
+			CD: core.Cooldown{
+				Timer:    warrior.NewTimer(),
+				Duration: time.Minute * 1,
+			},
+			ModifyCast: func(sim *core.Simulation, spell *core.Spell, cast *core.Cast) {
+				if warrior.AutoAttacks.MH().SwingSpeed == warrior.AutoAttacks.OH().SwingSpeed {
+					warrior.AutoAttacks.StopMeleeUntil(sim, sim.CurrentTime+cast.CastTime, true)
+				} else {
+					warrior.AutoAttacks.StopMeleeUntil(sim, sim.CurrentTime+cast.CastTime, false)
+				}
+			},
+			IgnoreHaste: true,
+		},
+		DamageMultiplier: 1,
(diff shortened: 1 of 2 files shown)


sim/warrior/heroic_leap.go
```diff
@@ -0,0 +1,50 @@
...
+
+func (warrior *Warrior) RegisterHeroicLeap() {
+
+	numHits := warrior.Env.GetNumTargets()
+	warrior.RegisterSpell(core.SpellConfig{
+		ActionID:       core.ActionID{SpellID: 6544},
+		SpellSchool:    core.SpellSchoolPhysical,
+		ProcMask:       core.ProcMaskMeleeMHSpecial,
+		Flags:          core.SpellFlagMeleeMetrics | core.SpellFlagAPL,
+		ClassSpellMask: SpellMaskHeroicLeap | SpellMaskSpecialAttack,
```
rosenrusinov, line 17: The class spell mask needs to unique and right now you're giving every single special attack bit to it. Leave only the HeroicLeap spell mask here and add the heroic leap to the specialattack masks if that is what you want to do

sim/warrior/heroic_leap.go
```diff
@@ -0,0 +1,50 @@
...
+		ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
+			baseDamage := 1 + 0.5*spell.MeleeAttackPower()
+			curTarget := target
+
+			for hitIndex := int32(0); hitIndex < numHits; hitIndex++ {
+				damageTarget := spell.CalcDamage(sim, curTarget, baseDamage, spell.OutcomeMeleeWeaponSpecialHitAndCrit)
+				spell.DealDamage(sim, damageTarget)
+
+				curTarget = sim.Environment.NextTargetUnit(curTarget)
+			}
```
rosenrusinov, line 47: You need to first calculate all damages and store them into results and after that deal them in a second loop

sim/warrior/heroic_leap.go
MatthewKyle: Okay, Most of the other Warrior spells are implemented this way.  If you prefer I can add to special attack

sim/warrior/heroic_leap.go
MatthewKyle: Okay, so just calc store in an array then deal the damage in a separate loop?  What is the logic here?  Sorry fairly new

rosenrusinov: Thats how it works in-game. If you have something that procs from damage it shouldnt proc between hits as it will mess with the calculates that are still to be done on the other targets.

sim/warrior/heroic_leap.go
rosenrusinov: Ohh sorry I see now how warriors implements the class masks. Weird but I guess it was more convenient this way. Sorry you can leave this as it is

sim/warrior/heroic_leap.go
```diff
@@ -0,0 +1,57 @@
...
+			for hitIndex := int32(0); hitIndex < numHits; hitIndex++ {
+				results[hitIndex] = spell.CalcDamage(sim, curTarget, baseDamage, spell.OutcomeMeleeWeaponSpecialHitAndCrit)
+				curTarget = sim.Environment.NextTargetUnit(curTarget)
+			}
+
+			curTarget = target
+			for hitIndex := int32(0); hitIndex < numHits; hitIndex++ {
+				spell.DealDamage(sim, results[hitIndex])
+				curTarget = sim.Environment.NextTargetUnit(curTarget)
+			}
```
rosenrusinov, line 53: You dont need to iterate over the target again. The results has everything already you can just remove those extra lines

MatthewKyle: Okay, yeah makes sense

PR #202, opened 2024-04-26
update hurricane procMask

zoristaken:
agility Darkmoon Card: Hurricane procMask is not set to the correct value making it not proc
Updated according to the implementation of the strength version of the trinket


rosenrusinov (approved):
Needs to merge master and update tests but otherwise good

PR #203, opened 2024-04-26
Enhupdates

foreverpsycotic:
Making FT correctly scale with ap instead of SP
Adding more efficient APL
Changing the default talents
changing the default glyphs
changing the p1 bis.


rosenrusinov (approved):
Just needs a test update

PR #207, opened 2024-04-26
Fix typo & add Mastery to Warrior display stats

1337LutZ:
I was playing around with the new Cata sim and noticed the warrrior sim was not showing Mastery.
- Added missing mastery in character display stats in the Sim UI for Warrior
- Fixed minor typo


1337LutZ:
@NerdEgghead Please double check if it's ok to add mastery as a epStat, and do we already have a default Ep value for mastery?

rosenrusinov (approved):

PR #221, opened 2024-04-27
[Warrior] Fixes

zoristaken:
Recklessness and Deadly Calm cannot be used together, also updated Inner Rage to use the same logic
Slam is affected by haste but the GCD is still 1.5. Would like some feedback on this part if possible.
Multiple spells were missing the IgnoreHaste flag, added the ones I could find.

### DIFF
--- a/sim/warrior/arms/deadly_calm.go
+++ b/sim/warrior/arms/deadly_calm.go
@@ -39,9 +39,6 @@ func (war *ArmsWarrior) RegisterDeadlyCalm() {
 		ClassSpellMask: warrior.SpellMaskDeadlyCalm,
 
 		Cast: core.CastConfig{
-			DefaultCast: core.Cast{
-				GCD: 0,
-			},
 			CD: core.Cooldown{
 				Timer:    war.NewTimer(),
 				Duration: time.Minute * 2,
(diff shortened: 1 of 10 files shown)


sim/warrior/arms/deadly_calm.go
```diff
@@ -40,21 +39,19 @@ func (war *ArmsWarrior) RegisterDeadlyCalm() {
...
 				Timer:    war.NewTimer(),
 				Duration: time.Minute * 2,
 			},
 		},
 		ProcMask: core.ProcMaskEmpty,
 		RageCost: core.RageCostOptions{Cost: 0},
+
 		ExtraCastCondition: func(sim *core.Simulation, target *core.Unit) bool {
-			return !war.HasActiveAuraWithTag(warrior.InnerRageExclusionTag)
+			return !war.InnerRageAura.IsActive() && !war.RecklessnessAura.IsActive()
```
Swillis57, line 51: Per behavior on the beta this should put Reck on CD for 10 seconds and Reck should put DC on CD for 12 seconds

rosenrusinov: You can use the Shared Cooldown field in the spells configuration. The way it works is to link a 2nd timer that can have a different duration and be activated from outside (in this case from the other ability of the 2)

rosenrusinov (approved):

PR #231, opened 2024-04-27
Feature/add armortype and primary stat for armor spec

1337LutZ:
I figured it would be nice to have the Armor Specialisation buff handled a bit more organized and hopefully less prone to forget.

- Added an ArmorType and PrimaryStat value on character level since:
  - Each class shares the same ArmorType
  - Some specs have a different primary stat but that is more often an exception than a norm
  - You set it once and then forget about it instead of applying it for each spec differently / with different checks in place
- The Armor Specialisation buff is now applied for each class except Druid (due to shapeshifting)
- Added a warning in the UI to tell someone if they are not benefitting from the Armor Specialisation buff
- Fixed Enhancement incorrectly being treated as Elemental (if else check) 
- Update Rogue & Enhancement shaman tests


rosenrusinov (changes requested):
There is no reason to store the primary stat or the armor type in the character struct when they are not used for anything so that change is pointless. Even more so that change makes the initialization of the armor specialization now require 2 distinctive changes in code when in reality it can't work without doing them both and only adds complexity for no gain.

I like the warning message so we can get that in tho.

1337LutZ:
> There is no reason to store the primary stat or the armor type in the character struct when they are not used for anything so that change is pointless. Even more so that change makes the initialization of the armor specialization now require 2 distinctive changes in code when in reality it can't work without doing them both and only adds complexity for no gain.
> 
> I like the warning message so we can get that in tho.

I've changed it back and added the logic for all the specs with the correct primary stats & armor types 🤞

rosenrusinov (approved):

PR #236, opened 2024-04-28
[Warrior] fixes mortal strike, slam, shattering glyph

MatthewKyle:
Based on PTR/Beta:
Changed mortal strike to 150% weapon damage
Changed Slam to 145%

Removed Minor Glyph of Shattering Throw  no longer exists in Cata


Swillis57 (approved):

sim/warrior/arms/mortal_strike.go
```diff
@@ -35,7 +35,7 @@ func (war *ArmsWarrior) RegisterMortalStrike() {
 		CritMultiplier:   war.DefaultMeleeCritMultiplier(),
 
 		ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
-			baseDamage := 423 + 0.8*(spell.Unit.MHNormalizedWeaponDamage(sim, spell.MeleeAttackPower()))
+			baseDamage := 634 + 1.5*(spell.Unit.MHNormalizedWeaponDamage(sim, spell.MeleeAttackPower()))
```
rosenrusinov, line 38: The 1.5 multiplier should also be applied to the base damage. How we normally model weapon skills is to put the multiplier in the spells `DamageMultiplier` field. The base damage should be the exact same base that you get for this spell using the `Class Base Scaling * Coefficient`. Mortal Strikes Coeff shows as `0.37599998713` and warriors base scaling value you can get from `core.GetClassSpellScalingCoefficient` it comes out at 423 at level 85 which is the correct value it had before your update

sim/warrior/slam.go
```diff
@@ -44,7 +44,7 @@ func (warrior *Warrior) RegisterSlamSpell() {
 
 		// TODO: check if the OH SMF hit is on a separate attack table roll
 		ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
-			baseDamage := 431 + 1.1*spell.Unit.MHWeaponDamage(sim, spell.MeleeAttackPower())
+			baseDamage := 625 + 1.25*spell.Unit.MHWeaponDamage(sim, spell.MeleeAttackPower())
```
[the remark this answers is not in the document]
rosenrusinov: Same comment as above. Fix how you apply the weapon damage % and udpate to the correct base value (using base scaling * coeff instead of a hardcoded precomputed number

proto/warrior.proto
```diff
@@ -113,7 +113,6 @@ enum WarriorMinorGlyph {
 	GlyphOfEnduringVictory = 43400;
 	GlyphOfFuriousSundering = 45793;
 	GlyphOfIntimidatingShout = 45794;
-	GlyphOfShatteringThrow = 206953;
```
rosenrusinov, line 116: Add this glyph id to the database overrides to remove it as selectable from the ui

rosenrusinov (changes requested):
Update weapon attacks with correct way to apply modifiers and base damage

rosenrusinov:
I pushed a change to master that generalizes how we setup the base class scaling You can update this PR with it. Warrior should refactor its base damage values to use SpellScaling * Coeff so its clearer from where the base damage comes and not just a magic number in the code

proto/warrior.proto
MatthewKyle: Changes Made.  I revert the glyph for now and now calculate the damage pct and extra damage for MS and Slam

sim/core/spell.go
```diff
@@ -55,9 +55,10 @@ type SpellConfig struct {
 	ExpectedInitialDamage ExpectedDamageCalculator
 	ExpectedTickDamage    ExpectedDamageCalculator
 
-	Dot    DotConfig
-	Hot    DotConfig
-	Shield ShieldConfig
+	Dot                DotConfig
+	Hot                DotConfig
+	Shield             ShieldConfig
+	SpellScalingConfig SpellScalingConfig
```
rosenrusinov, line 61: Remove this line from here as it seems you added it as a mistake.

sim/core/spell_scaling_config.go
```diff
@@ -0,0 +1,19 @@
...
+}
+
+func (config *SpellScalingConfig) CalcSpellDamagePct() float64 {
+
+	return config.BaseWeapon_Pct + (config.EffectPerLevel * (float64(config.MaxSpellLevel - config.BaseSpellLevel)) / 100)
+}
+
+func (config *SpellScalingConfig) CalcAddedSpellDamage() float64 {
+	return config.Coefficient * config.ClassSpellScaling
+}
```
rosenrusinov, line 19: This is specific to the Spell Effect for `WEAPON_PERCENT_DAMAGE` and should probably not be here or if you want it to be part of core name it something that is clear its for this effect and not general `SpellScalingConfig`. You can move it to the warrior package if not.

MatthewKyle: Moved to Warrior Package and renamed it for clarity.

rosenrusinov (approved):

PR #240, opened 2024-04-28
Fixes #213 - Add minimum item count and fix targets

### LINKED ISSUE #213: Can't add a target to encounters
wowsims error
**Describe the bug**
So I accidentally deleted the target for Encounter, now there's no option to add it back and I get the error message provided in the ss. There's no way to add a new target either.

**To Reproduce**
1. Go to Encounter
2. Click on Advanced and then delete all current targets and try to sim again. It will provide the error message and no way to add a new target.

**Expected behavior**
The advanced tab should not disappear and allow you to add a new target.

**Sim Links and Screenshots**
wowsims
wowsims error

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser Chrome

**Smartphone (please complete the following information):**

**Additional context**

1337LutZ:
Fixes #213 
- Added `minimumItems` prop to the ListPicker enforce a minimum amount of items in the ListPicker to prevent having 0 items if it's not supported
  - Hides delete button on `n`entries if `minimumItems` is set
- Targets always expects to at least have 1 entry, which also makes sense. Since there's no reason to have 0 targets (that I can think of), since a user still has to edit it anyways, even when adding a new entry I believed that just making sure you can never has no targets would be an appropriate solution.

### DIFF
--- a/ui/core/components/encounter_picker.ts
+++ b/ui/core/components/encounter_picker.ts
@@ -184,6 +184,7 @@ class AdvancedEncounterModal extends BaseModal {
 				index: number,
 				config: ListItemPickerConfig<Encounter, TargetProto>,
 			) => new TargetPicker(parent, encounter, index, config),
+			minimumItems: 1,
 		});
 	}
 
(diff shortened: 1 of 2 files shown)


rosenrusinov (approved):

ui/core/components/encounter_picker.ts
```diff
@@ -689,10 +690,10 @@ function makeTargetInputsPicker(parent: HTMLElement, encounter: Encounter, targe
 }
 
 function equalTargetsIgnoreInputs(target1: TargetProto | undefined, target2: TargetProto | undefined): boolean {
-	if ((target1 == null) != (target2 == null)) {
```
1337LutZ, line 692: This check was broken by the way, since target could be `undefined` as well.

PR #244, opened 2024-04-28
[Shaman] Switch to using ClassSpellScaling and added Earth Elemental

Polynomix:
Requested change to SpellScaling(tests result updated because of the precision differences)
Added Earth Elemental
Added Earth elemental totem and Fire elemental totem to the call of the element picker.


sim/shaman/chain_lightning.go
```diff
@@ -41,6 +41,7 @@ func (shaman *Shaman) newChainLightningSpell(isElementalOverload bool) *core.Spe
 		for hitIndex := int32(0); hitIndex < numHits; hitIndex++ {
 			totalDamage := baseDamage * bounceReduction
 
+			//TODO : damage reduction from glyph need to be applied to spell power part of the damage also
```
rosenrusinov, line 44: To achieve this TODO instead of `baseDamage * bounceReduction` do `spell.DamageMultiplier *= bounceReduction` in the calc loop then do `spell.DamageMultiplier /= bounceReduction` in the deal loop

sim/shaman/earth_elemental_pet.go
```diff
@@ -0,0 +1,105 @@
...
+func (earthElemental *EarthElemental) Initialize() {
+
+}
+
+func (earthElemental *EarthElemental) Reset(_ *core.Simulation) {
+
+}
+
+func (earthElemental *EarthElemental) ExecuteCustomRotation(sim *core.Simulation) {
+	earthElemental.ExtendGCDUntil(sim, sim.CurrentTime+time.Second)
```
rosenrusinov, line 76: Dont need to call this just an empty method should do

rosenrusinov (approved):

PR #249, opened 2024-04-29
[UI] Add quick swapping (Gems / Enchants)

1337LutZ:
Simming brings a lot of swaps between gems and enchants. I decided to make this easier for myself and make enchants/gems much quicker to swap around:
- Hover opens quick swap interface and click opens detailed view
- Added wowhead tooltip to gems
- **Enchants**
  - Favorite enchants will be shown as options
  - Options will be filtered per item slot
- **Gems**
  - Favorite gems will be shown as options
  - Options are sorted by color
  - Meta gem is only visible in Head slot
- No Favorites eligible/chosen will show a message
- Button in footer opens the detailed modal
- Due to Popper being lackluster in support for interactive tooltips, and we still had Tippy I replaced every Bootstrap Popover/Tooltip with Tippy.
- Show quick swap option is behind a user setting and disabled on mobile
  
image
image
image


1337LutZ:
**Notes from Discord**
- Add separate setting to enable favorite quick swap
- Add hover / click for the Popover / Default behavior

1337LutZ:
> **Notes from Discord**
> 
> * Add separate setting to enable favorite quick swap
> * Add hover / click for the Popover / Default behavior

Fixed and edited OP

kayla-glick (approved):
✨

PR #259, opened 2024-04-29
Gear picker improvements

kayla-glick:
Adds the ability to swap between different item slots from within the gear picker. Improves performance a bit by using a singleton modal that dynamically updates as you select different slots, rather than initializing a new modal every time. Also updates the reforge tab to use the conventions of other tabs

image
image


kayla-glick:
Leaving off for now because I'm going out of town. I ran into issues with the reforging tab so I've been reworking that a bit and trying to get it to use the existing tab framework that other tabs use. I have it displaying reforge data like so:

image

but I'm running into a couple of issues that need to be addressed still:

- Changing to one of the other tabs causes the reforge stats to disappear and I don't know why yet
- The first time the modal is opened it correctly attaches within the SimUI so it gets the inherited styles. However closing it and re-opening it causes the modal to attach to the `<body>` instead which is not what we want
- Can't currently select a reforge and I'm not sure why yet

1337LutZ:
- **Changing to one of the other tabs causes the reforge stats to disappear and I don't know why yet**
  - This was because a old `.reforging` parameter wasn't changed yet 
- **The first time the modal is opened it correctly attaches within the SimUI so it gets the inherited styles. However closing it and re-opening it causes the modal to attach to the <body> instead which is not what we want**
  - I added a boolean that allows modals to not automatically be disposed when closed. Not sure if this was the right fix, but for these modals it seemed not necessary to keep re-initializing them.
- **Can't currently select a reforge and I'm not sure why yet**
  - Same as the first problem + there was a check that was broken here: https://github.com/wowsims/cata/pull/259/commits/72449be91496398239c14182576a35aa10fbed58#diff-fba6d1d7b1912ba0a322a9c672fe5b06eba50d893f4ec5063e022de71ad7aa00R232 and here https://github.com/wowsims/cata/pull/259/commits/72449be91496398239c14182576a35aa10fbed58#diff-c6ed59d44ea73966b3d3a7d8a23e15fbad7fb5f6cd6323258c6c90d8cc69e6fdR111
- I also made the Glyph picker 1 modal, since it was initializing 9 modals.

ui/core/components/gear_picker/gear_picker.tsx
```diff
@@ -1126,8 +1126,9 @@ export class ItemList<T extends ItemListType> {
 		makePhaseSelector(this.tabContent.getElementsByClassName('selector-modal-phase-selector')[0] as HTMLElement, player.sim);
 
 		if (label == 'Items') {
```
1337LutZ, line 1128: This can use the `SelectorModalTabs.Items`

ui/core/components/gear_picker/gear_picker.tsx
```diff
@@ -1088,20 +1047,27 @@ export class ItemList<T> {
...
-		this.slot = slot;
+		this.slot = currentSlot;
 		this.gearData = gearData;
 		this.currentFilters = this.player.sim.getFilters();
 
 		const tabContentId = (label + '-tab').split(' ').join('');
-		const selected = label === config.selectedTab;
+		const selected = label === currentTab;
+
+		let itemLabel = 'Item';
```
1337LutZ, line 1057: Maybe save these as enums like the `SelectorModalTabs`?

ui/core/components/gear_picker/gear_picker.tsx
```diff
@@ -1138,7 +1106,12 @@ export class ItemList<T> {
 
 		makeShow1hWeaponsSelector(this.tabContent.getElementsByClassName('selector-modal-show-1h-weapons')[0] as HTMLElement, player.sim);
 		makeShow2hWeaponsSelector(this.tabContent.getElementsByClassName('selector-modal-show-2h-weapons')[0] as HTMLElement, player.sim);
-		if (!(label == 'Items' && (slot == ItemSlot.ItemSlotMainHand || (slot == ItemSlot.ItemSlotOffHand && player.getClass() == Class.ClassWarrior)))) {
+		if (
+			!(
+				label == 'Items' &&
```
1337LutZ, line 1111: This can use the `SelectorModalTabs.Items`

ui/core/components/gear_picker/gear_picker.tsx
1337LutZ, line 1: This is a personal thing, but could we change all those: `.getElementsByClassName()[0] as XXX` to `.querySelector<HTMLElement>('.selector')!` ? That way we dont have to typecast, pass a generic and just tell TS that it's always defined with the `!`

ui/core/components/gear_picker/gear_picker.tsx
kayla-glick: Well these are `Item` for everything except for Reforge. Originally I had another case with more that I changed so probably doesn't need a switch stateent

PR #263, opened 2024-04-29
Arcane Mage updates

Qingle:
- Made Arcane Blast mana scaling Pct
- Removed mana scaling from Arcane Explosion from Arcane Blast debuff
- Fixed Arcane Blast & Arcane Explosion mana costs
- Fixed Arcane Potency being consumed by non-damaging spells

### DIFF
--- a/sim/mage/talents_arcane.go
+++ b/sim/mage/talents_arcane.go
@@ -238,6 +238,7 @@ func (mage *Mage) applyArcanePotency() {
 	})
 
 	var procTime time.Duration
+	const ArcanePotencyClassMask = MageSpellArcaneBlast | MageSpellArcaneBarrage | MageSpellArcaneMissilesCast | MageSpellArcaneExplosion
 	mage.ArcanePotencyAura = mage.RegisterAura(core.Aura{
 		Label:     "Arcane Potency",
 		ActionID:  core.ActionID{SpellID: 57531},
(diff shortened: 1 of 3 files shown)


sim/mage/talents_arcane.go
```diff
@@ -257,6 +258,11 @@ func (mage *Mage) applyArcanePotency() {
 			if sim.CurrentTime == procTime {
 				return
 			}
+
+			if spell.ClassSpellMask&ArcanePotencyClassMask == 0 {
```
InDebt, line 262: It might seem unreasonable to cast anything but those spells, but this shoudl be the same SpellMask used for the SpellMod. Else Fireball / Frostbold and other spells will be affected but do not consume this effect.

In general this spell's interaction with ArcaneMissles as to be reworked. Every tick of AM can proc clear cast and refresh AP but after the spell is done the crit is consumed which should not happen. AP is consumed on channel start.

AM snapshots the crit value on cast (only crit) and uses that cit value for every tick spell. So this logic should be implemented in AM not in AP.

Qingle: That's quite a hefty list from what I understand, is there any other way to link the two or have a differential for the damage spells over utility?

Qingle: Probably need to do a bit more testing as to what interacts with it before doing that. Flame Orb for example doesn't proc this but procs clearcasting/AM so it's inconsistent.

PR #271, opened 2024-04-30
[UI] Add fix for phase change filter when 0

### LINKED ISSUE #247: Filters do not apply correctly/at all when viewing sources of items
**Describe the bug**
When selecting sources of items to filter, the filters do nothing

**To Reproduce**
1. Click on any item, e.g., helmet
2. Click filters
3. Mark/unmark sources as needed

**Expected behavior**
The underlying list of items should be filtered based on your selected sources, but they're not

**Sim Links and Screenshots**
image

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Chrome, tested in Firefox as well
 - Version: Unsure

1337LutZ:
Partially fixes #247 
- Pre-patch is phase 0, thus the `> 0` check had to be removed
- Other filters don't work since `sources` is populated by AtlasLootClassic and there's no Cataclysm dataset yet.

### DIFF
--- a/ui/core/components/other_inputs.ts
+++ b/ui/core/components/other_inputs.ts
@@ -68,7 +68,6 @@ export function makePhaseSelector(parent: HTMLElement, sim: Sim): EnumPicker<Sim
 	return new EnumPicker<Sim>(parent, sim, {
 		extraCssClasses: ['phase-selector'],
 		values: [
-			{ name: 'Prepatch', value: 0 },
 			{ name: 'Phase 1 (Cataclysm)', value: 1 }
 		],
 		changedEvent: (sim: Sim) => sim.phaseChangeEmitter,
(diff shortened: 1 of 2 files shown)


rosenrusinov:
We dont support pre-patch and never will in the backend so enabling it on the front end will only confuse users

1337LutZ:
> We dont support pre-patch and never will in the backend so enabling it on the front end will only confuse users

Hmm, but the item filter is there, but just broken. Or should I just remove it instead?
image

rosenrusinov:
I think we're better off with removing the prepatch filter yes. It wont work anyway as the characters simmed will still be 85

1337LutZ:
> I think we're better off with removing the prepatch filter yes. It wont work anyway as the characters simmed will still be 85

Resolved 👍

PR #291, opened 2024-05-01
Port and extend movement APL from SoD

InDebt:
Sooo I kind of ported, yet half rewrote the movement stuff.
In cata auto shots still work while moving, and melee hits should not be disabled either while moving, except when you're out of range so I changed that behaviour a bit and updated the way movement updates are calculated. In the previous implementation, spells could have been casted with invalid range constraints as the movement was updated only every 500 ms. The impact wasn't big as the player was moving with 2yards/s but with 7 yards this was more impactful.

## What's included?
* APL to check if a player is moving
* APL to move for a specified amount of time (no distance change)
* APL to move to a specified distance
* Added movement hooks - which I ended up not used right now 
  * Thought i'd use them for shadow but TriggerAuras with dynamic proc chances do not play nice together
* In theory added mob movement (just added the correct speed for mobs)
* Added support for movement for Shadow APL and added Hunter Apsect of the Fox
* Added SpellMod that adds `Allow casting while moving` for Aspect
* Added Min/Max Range for Autoshot + Melee Attacks for now
  * If need a general system for min / max range checks and spells we can add that later
* Changed the default test range to 0m and added a default for ranged (hunter) to 5.1 meter outside min range
  * Right now melees were at 30m by default which seemed to be a damage loss 😁

Should replace PR #34 

## Open Topics
* Shoudl we provide a `Distance from Target` APL value
  * Didn't miss it for Shadow but not sure about other cases


proto/apl.proto
```diff
@@ -70,16 +70,18 @@ message APLAction {
...
+		APLActionMove move = 18;
+        APLActionMoveDuration move_duration = 22;
 
         // Class or Spec-specific actions
-        APLActionCatOptimalRotationAction cat_optimal_rotation_action = 18;
+        APLActionCatOptimalRotationAction cat_optimal_rotation_action = 19;
 
         // Internal use only, not exposed in UI.
-        APLActionCustomRotation custom_rotation = 19;
+        APLActionCustomRotation custom_rotation = 21;
```
rosenrusinov, line 80: Changing proto indices is a bad patter as it breaks any saved data of users. Always add new actions to the next index without changing other actions/values indices

sim/core/attack.go
```diff
@@ -409,6 +418,16 @@ func (unit *Unit) EnableAutoAttacks(agent Agent, options AutoAttackOptions) {
...
+		ExtraCastCondition: func(sim *Simulation, target *Unit) bool {
+			// ranged has minimum range of 5
+			return unit.DistanceFromTarget > 5
+		},
+	}
+
+	// set character at ranged if primary attack is ranged
+	if options.AutoSwingRanged && unit.DistanceFromTarget == 0 {
+		unit.DistanceFromTarget = 5.1
 	}
```
rosenrusinov, line 431: I dont think this should be set here. You can overwrite the current distance from target from just disabling/enabling auto attacks now. This default should be probably set per spec during initialization of the spec

rosenrusinov: Also thats why we have to Move to Distance action supporting prepull. So users can set their prefered starting distance during prepull

sim/core/movement.go
```diff
@@ -0,0 +1,139 @@
...
+		ApplyEffects: func(sim *Simulation, target *Unit, spell *Spell) {
+			unit.moveAura.Activate(sim)
+			unit.moveAura.SetStacks(sim, max(int32(unit.DistanceFromTarget), 1))
+		},
+	})
+
+	unit.RegisterResetEffect(func(s *Simulation) {
+		unit.Moving = false
+		unit.movementAction = nil
+	})
```
rosenrusinov, line 49: You can make this be part of the moveAura OnExpire maybe so we dont add extra reset effects?

sim/core/unit.go
```diff
@@ -686,3 +693,16 @@ func (unit *Unit) GetTotalAvoidanceChance(atkTable *AttackTable) float64 {
 	block := unit.GetTotalBlockChanceAsDefender(atkTable)
 	return miss + dodge + parry + block
 }
+
+func (unit *Unit) MultiplyMovementSpeed(sim *Simulation, amount float64) {
+	unit.PseudoStats.MovementSpeedMultiplier *= amount
+}
```
rosenrusinov, line 699: This should check for any movement currently happening and updated timers/durations based on new speed

sim/core/attack.go
InDebt: Oh missed that. Though this method would only be called once on sim init. Will have a look to get that done more properly

ToxicKevinFerm: Another thing is it's gonna be pretty ugly to add
```
		ExtraCastCondition: func(sim *Simulation, target *Unit) bool {
			// ranged has minimum range of 5
			return unit.DistanceFromTarget > 5
		},
```
To every Hunter spell. I'd want both MinRange and MaxRange to be part of the SpellConfig if we're gonna have movement and boss api imo, much cleaner

sim/core/movement.go
```diff
@@ -0,0 +1,139 @@
...
+	movementAction := MovementAction{
+		startTime:   sim.CurrentTime,
+		speed:       speed,
+		srcPosition: unit.DistanceFromTarget,
+	}
+
+	movementAction.NextActionAt = endTime
+	movementAction.OnAction = func(sim *Simulation) {
+		unit.FinalizeMovement(sim)
+	}
```
rosenrusinov, line 114: So if I'm understanding this correct you only update the distance from target at the start of the movement and at the end of the movement. What happens if I'm at 3 yards and want to move to 8 yards. Will my melee auto attack get disabled when I cross the 5 yards threshold or will it get instantly disabled when i start moving?

rosenrusinov: Ok, i see that you are updating the current position from the APLs callback which also works (maybe even better so every action has the precise current position)

InDebt: Distance is updated before APL actions if the unit is moving to ensure correct positions. Due to the extra conditions in the Attacks if we're in range we attack, if not - we don't. Melee ranged attack will always happen on retail when you're within range limitations.

Given your comment below, I can also disable enable them based on distance during movement udpate.

sim/core/attack.go
rosenrusinov: > Oh missed that. Though this method would only be called once on sim init. Will have a look to get that done more properly

This method actually can be called during sim as well (some spells disable/enable autoattacks)

sim/core/spell.go
```diff
@@ -468,6 +468,14 @@ func (spell *Spell) CanCast(sim *Simulation, target *Unit) bool {
 		return false
 	}
 
+	// While moving only instant casts are possible
+	if spell.Flags&SpellFlagCanCastWhileMoving == 0 && spell.DefaultCast.CastTime > 0 && spell.Unit.Moving {
```
rosenrusinov, line 472: This will return the cast time without the multiplier and without any modify casts being applied. The actual cast will use them so its not that big of a deal but we should fix this in a future pr

rosenrusinov (approved):

sim/core/spell.go
InDebt: Hm.. might be something we want to add into SOD branch as well then as I just copied this part over and added the extra check for the SpellFlag. So SoD is not using the CastMultiplier either right now

PR #295, opened 2024-05-02
[Warrior] Added Charge, updated defaults arms

MatthewKyle:
Added Charge and modified by blitz
Updated Gear. APL., Settings

updated tests

fixed test outputs


sim/warrior/arms/talents.go
```diff
@@ -59,6 +60,7 @@ func (war *ArmsWarrior) applyTasteForBlood() {
 		ClassSpellMask: warrior.SpellMaskRend,
 		ICD:            5 * time.Second,
 		ProcChance:     procChance,
+		SpellFlags:     core.SpellFlagAPL,
```
rosenrusinov, line 63: You can't have triggers as APL spells. APL spells are spells that the agent can cast from the APL Cast action

sim/warrior/arms/talents.go
```diff
@@ -50,6 +50,7 @@ func (war *ArmsWarrior) applyTasteForBlood() {
 		ActionID: actionID,
 		Duration: time.Second * 9,
 		Tag:      warrior.EnableOverpowerTag,
+		Icd:      &core.Cooldown{Duration: time.Second * 6},
```
rosenrusinov, line 53: this ICD is doing nothing atm. In the handler for activating this aura you need to check if the ICD is ready then activate and use the ICD

sim/warrior/charge.go
```diff
@@ -0,0 +1,33 @@
+package warrior
+
+import (
+	"time"
+
+	"github.com/wowsims/cata/sim/core"
+)
+
+func (warrior *Warrior) RegisterCharge() {
```
rosenrusinov, line 9: Can charge be used in combat and with no minimum range now?

sim/warrior/charge.go
```diff
@@ -0,0 +1,33 @@
...
+	metrics := warrior.NewRageMetrics(core.ActionID{SpellID: 100})
+
+	warrior.Charge = warrior.RegisterSpell(core.SpellConfig{
+		ActionID:       core.ActionID{SpellID: 100},
+		SpellSchool:    core.SpellSchoolPhysical,
+		Flags:          core.SpellFlagAPL,
+		ClassSpellMask: SpellMaskCharge,
```
rosenrusinov, line 16: missing ProcMask

rosenrusinov (changes requested):
Cant mark non castable spells as APL

sim/warrior/arms/talents.go
```diff
@@ -59,6 +60,7 @@ func (war *ArmsWarrior) applyTasteForBlood() {
 		ClassSpellMask: warrior.SpellMaskRend,
 		ICD:            5 * time.Second,
```
rosenrusinov, line 61: This ICD is 5s for the proc but the application aura is 6s. This looks weird to me

sim/warrior/arms/talents.go
MatthewKyle: Yeah, this is only for tracking the auras cooldown but required the SpellAPLFlag on the Trigger to use.  Removiung till I figure something better out.

sim/warrior/charge.go
MatthewKyle: Juggernaut makes Charge and Intercept share a CD.  I did not add intercept as there is no DPS use currently.

sim/warrior/arms/talents.go
MatthewKyle: Removed.  I was trying to make the proc ICD visible.

PR #324, opened 2024-05-05
Concurrent raid sim

FelixPflaum:
Starting as a little experiment this turned out to actually be a viable way to add concurrency. 

It's achieved by spreading iterations over multiple sims, which run in their own go routine. Results are then combined, both for progress updates and the final RaidSimResult. This is entirely transparent and only requires the replacement of `go RunSim(request, progress) `with the new `RunConcurrentRaidSimAsync(request, progress)` to be used.
Edit: Sadly it now also requires a minor change to protos to properly get stdevs from partial sim results.

There's no impact on the web version because wasm is single thread anyway, therefore I made it not even compile for wasm.

It currently has test cases for MM Hunter, Blood DK and cat Feral, comparing the entire results of a normal single threaded sim with those of a multi threaded one. This should catch all possible errors introduces by the result combination. Additionally the previously added leak test uses concurrency now, which should make sure of everything working even more.

I'd like some feedback especially on how to implement it. Could be as an optional experimental feature, be it via a checkbox in the UI (this would require more code changes though) or via a compile flag. Or it could just be made the default for the standalone binary, maybe after some testing by multiple people.

### Minor imperfections
- CastTimeMs in TargetedActionMetrics can be off by 1ms per thread due to it being rounded.
- There's some float tolerance in tests, mainly for resource metrics and stdevs. 
- If a thread errors other threads continue. Though, progress is ignored and only the first error is passed to the client.

The differences between single and multi thread results shouldn't matter anywhere though.


FelixPflaum:
Turns out hps for the DK test case is off
`HPS expected 4209.433 but was 4210.801 for multi threaded sim!`
All other values are the same (or within the 0.0001 tolerance) for the now 3 test specs.

1337LutZ:
I am vouching for just implementing this as a default for all local native sims. There's no real reason you ever would not want free performance.

FelixPflaum:
Result mismatches are most likely all caused by the sim not properly resetting everything between iterations. See https://github.com/wowsims/cata/pull/326

PR #325, opened 2024-05-06
[Warlock] Couple fixes and updates

ghronkrepps:
- remove fel armor option, add it as standard spellpower add for all warlock specs
- add shadowflame
- add soulburn and effect on pet summon and soulfire
- fix shadowburn damage and also requires by chaos bolt talent
- remove a couple comments
- set felhunter summon pet to change to the correct pet


Tharre:
You need to set the values for demon armor to reserved in the .proto, otherwise you'll be breaking all the existent configs.

Also are we actually sure that removing that option is a good idea? Might make testing a little bit harder and it's not like the option for it is taking up much space in the UI.

rosenrusinov:
Ay spell that can be cast from a repull action has no place in the ui. Are the warlock armors static? If you never change the armor and they never expire then you can remove it from the apl and make it an option but if its not a static buff it should be removed from the options and be handled by the apl like any other spell

Tharre:
They are static; they don't expire and give a fixed amount of stats. I don't think there's ever a situation where you'd swap them in combat, so having them in the APL is indeed pointless.

ghronkrepps:
> You need to set the values for demon armor to reserved in the .proto, otherwise you'll be breaking all the existent configs.
> 
> Also are we actually sure that removing that option is a good idea? Might make testing a little bit harder and it's not like the option for it is taking up much space in the UI.

Hmm I asked in the warlock channel if I should change it to "always on" and didn't get much response other than "we will never use demon armor"

ghronkrepps:
As a user I don't know why you would want this as an option considering all of the points you just made about it. It's a self buff that will be up 100% of the time unless you forget it. :D

sim/warlock/warlock.go
```diff
@@ -77,6 +80,7 @@ func (warlock *Warlock) Initialize() {
 
 	// base scaling value for a level 85 warlock
 	warlock.ScalingBaseDamage = 962.335630
+	warlock.SoulShards = 4
```
rosenrusinov, line 83: This should be done in the Reset method to make sure it sets it to 4 at start of next iterations as well

rosenrusinov:
> As a user I don't know why you would want this as an option considering all of the points you just made about it. It's a self buff that will be up 100% of the time unless you forget it. :D

This is also fine but how do you communicate this to the user? I think you should atleast add a permanent dummy aura with the fel armor id so users can see that the buff is active in the buffs tab

PR #342, opened 2024-05-07
Ele Shaman Updates

raethkcj:
Add Fulmination split metrics and improve Fire Ele's base stats, base damage, and coefficients

### DIFF
--- a/sim/shaman/fire_elemental_spells.go
+++ b/sim/shaman/fire_elemental_spells.go
@@ -8,7 +8,7 @@ import (
 
 func (fireElemental *FireElemental) registerFireBlast() {
 	fireElemental.FireBlast = fireElemental.RegisterSpell(core.SpellConfig{
-		ActionID:    core.ActionID{SpellID: 13339},
+		ActionID:    core.ActionID{SpellID: 57984},
 		SpellSchool: core.SpellSchoolFire,
 		ProcMask:    core.ProcMaskSpellDamage,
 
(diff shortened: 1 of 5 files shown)


sim/shaman/fire_elemental_spells.go
```diff
@@ -61,18 +61,19 @@ func (fireElemental *FireElemental) registerFireNova() {
 		DamageMultiplier: 1,
 		CritMultiplier:   2, //Estimated from beta testing
 		ThreatMultiplier: 1,
+		BonusCoefficient: 1.00,
 
 		ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
 			for _, aoeTarget := range sim.Encounter.TargetUnits {
-				baseDamage := sim.Roll(1527, 1731) * sim.Encounter.AOECapMultiplier() //Estimated from beta testing
+				baseDamage := sim.Roll(1370, 1574) * sim.Encounter.AOECapMultiplier() //Estimated from beta testing
```
Polynomix, line 68: Looks like you used 157 SP to get to the new value, i'm not sure if fire ele also have the first 10 int not giving spell power or not

Polynomix: now that i think about it, it should adds up to the same damage when inheriting stats from the owner since you get back the 10 SP that you previously lost because of no base int

Polynomix: yes

PR #346, opened 2024-05-07
Draft: Proof of concept - WASM concurrency

1337LutZ:
I've made a PoC to start {n} web workers and achieve the same as the local sim. It works pretty well and the results are pretty nice as well: 

Concurrent (my cpu = 20) WASM 100k: 
image

Single WASM 100k: 
image

I have little to no Go knowledge so I don't know how to properly make the Go code to create combined raid sim results get converted to JS/TS. However, as a proof of concept this works.

- [x] Create workers per thread available
- [x] Merge and display combined sim results 
- [x] Add `workerType` in the `ready` callback to determine if  it should use the web concurrency or spawn 1 worker locally since it handles it in Go
- [x] Cleanup some duplicated code 
- [ ] Add setting for managing {n} threads
- [ ] Re-use the Go sim result sharing logic (Need help with this)
- [ ] Add tests (?)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,8 +9,9 @@
             "version": "0.1.0",
             "dependencies": {
                 "@popperjs/core": "^2.11.6",
+                "@types/pako": "^2.0.0",
                 "bootstrap": "^5.3.0",
-                "pako": "^1.0.11",
+                "pako": "^2.0.4",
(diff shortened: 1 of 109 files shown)


1337LutZ:
Closed since #463  is almost ready

PR #347, opened 2024-05-07
Several gear picker issues and styles

kayla-glick:
- Fixed issues with event handlers not being properly disposed of. We have to locally assign the functions then add callbacks to remove them instead of the previous approach.
  - This also fixed an issue where re-opening the gear picker caused the up and down arrow functions to run multiple times, leading to "skipping" slots
- Fixed an issue where switching between slots with the up and down arrows caused newly equipped gear to be equipped in the original slot that was opened
  - This required changing how the tab switching was handled by going through the actual item picker so that `gearData` is built correctly
- Fixed some mobile style issues (see screenshot)
  - Item icons were smaller than the bordered frame
  - There were some spacing and alignment issues on the thinnest screens
  - Adjusted XS screen padding of the main sim content to align with the sidebar padding

image

### DIFF
--- a/ui/scss/core/sim_ui/_main.scss
+++ b/ui/scss/core/sim_ui/_main.scss
@@ -1,42 +1,64 @@
 .sim-content {
-  min-width: calc(275px + 1vw);
-  display: flex;
-  flex-flow: column;
+	min-width: calc(275px + 1vw);
+	display: flex;
+	flex-flow: column;
 
-  .sim-main {
-    height: 80%;
-    display: flex;
-    flex-grow: 1;
+	.sim-main {
+		height: 80%;
+		display: flex;
+		flex-grow: 1;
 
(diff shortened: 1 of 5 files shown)


ui/scss/core/sim_ui/_main.scss
```diff
@@ -1,42 +1,64 @@
...
-        padding-top: $gap-width-sm * 2;
-        padding-bottom: $gap-width-sm * 2;
-      }
-    }
-  }
+	.sim-sidebar {
+		.sim-sidebar-content {
+			padding: $gap-width-sm * 2 $gap-width-sm;
+		}
+	}
```
kayla-glick, line 63: I feel like at some point we need to just blow away a bunch of CSS and rewrite it 🙃

1337LutZ (approved):

PR #348, opened 2024-05-07
Balance druid implementation

stn-ms:
some bug fixes to stuff that was reported to me directly.

Also split Moonfire into a dot spell and impact spell and added Sunfire. If one is already present, the other one should overwrite it.
Moved a few of the glyphs to a new glyphs.go
Added mastery bonuses to eclipses.
Insect Swarm ticks not critting.
A more advanced APL (thanks Shantie <3)


sim/druid/eclipse.go
```diff
@@ -56,15 +56,29 @@ func (druid *Druid) EnableEclipseBar() {
 }
 
 func (druid *Druid) RegisterEclipseAuras() {
+	masteryBonus := 0.16 + druid.GetMasteryPoints()*0.02
```
rosenrusinov, line 59: You dont have mastery fully initialized during registration of spells/auras. You need to update the value again in the auras OnGain methods before activating the mods. Also you should register to the On Mastery Changed callback to update the spell mods values when mastery changes due to procs

sim/druid/moonfire.go
```diff
@@ -43,29 +31,63 @@ func (druid *Druid) registerMoonfireSpell() {
...
 			},
 			OnTick: func(sim *core.Simulation, target *core.Unit, dot *core.Dot) {
 				dot.CalcAndDealPeriodicSnapshotDamage(sim, target, dot.OutcomeSnapshotCrit)
 			},
 		},
 
+		ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
+			result := spell.CalcOutcome(sim, target, spell.OutcomeAlwaysHit)
+
+			if result.Landed() {
```
rosenrusinov, line 48: no point of this if when the outcome is always hit

ui/core/proto_utils/action_id.ts
```diff
@@ -378,6 +378,16 @@ export class ActionId {
 				break;
 			case 'Fulmination':
 				name += ` (${this.tag + 3})`
+			case 'Moonfire':
+				if (this.tag == 1) {
+					name += ' (DoT)'
+				}
+				break;
```
rosenrusinov, line 385: you can have the switch case falltrough to the next case so just putting:
```
case 'Moonfire':
case 'Sunfire':
  dot...
```

rosenrusinov (commented):
Only real issue is the mastery related spell mods. Other things are nits

sim/druid/eclipse.go
stn-ms: I can move it there, but I don't think I need to update it as Eclipse is supposedly snapshotting mastery.

ui/core/proto_utils/action_id.ts
stn-ms: oh true, thanks

sim/druid/eclipse.go
rosenrusinov: the spell mod should still be properly updated on mastery changes. The dot will still properly snapshot the damage on application but you still have to have the correct value to snapshot from.

stn-ms: yep, that's done.

PR #350, opened 2024-05-07
Refactor media queries

1337LutZ:


ui/core/components/icon_enum_picker.tsx
```diff
@@ -63,10 +63,10 @@ export class IconEnumPicker<ModObject, T> extends Input<ModObject, T> {
 		this.config.changedEvent(this.modObject).on(_ => {
 			if (this.showWhen()) {
 				this.restoreValue();
-				this.rootElem.classList.remove('hide');
+				parent.classList.remove('hide');
```
1337LutZ, line 66: I added this because the parent element would still push away in flex/grid settings. Moving this to parent actually excludes it from css grid calcs.

PR #373, opened 2024-05-08
[Proposal] Sim abort controller

1337LutZ:
Currently there's no option to cancel pending/ongoing sims. 
There's a few scenario's where this can be a problem:
- Large amount of iterations and you want to change something on the fly.
- Accidental interactions with the Sim button. 

Right now a new sim is triggered for each time you press, even if the previous one hasn't finished. (We could also opt to disable the Sim button whilst a sim is going on)

**Local**
The server will run the simulation regardless and add it to the queue, whilst the application can only host 1 Sim request at a time.

**WASM** 
Currently you can reload the application to stop ongoing simulations.
Terminating a wasm worker should work and cancel the pending simulation on that worker (since we destroy it and wasm lives in the Worker), in contrary to the local version.

**Todo**

- [x]  Implement abort controller for sim requests
- [ ] Add a abort signal that the Go code can read. (afaik we keep track of the processes, but there's no way to actually abort them?)

### DIFF
--- a/ui/core/sim.ts
+++ b/ui/core/sim.ts
@@ -60,6 +60,7 @@ export enum SimSettingCategories {
 // Core Sim module which deals only with api types, no UI-related stuff.
 export class Sim {
 	private readonly workerPool: WorkerPool;
+	private readonly abortControllers: Record<string, AbortController> = {};
 
 	private iterations = 3000;
 	private phase: number = OtherConstants.CURRENT_PHASE;
@@ -241,7 +242,10 @@ export class Sim {
 		this.bulkSimStartEmitter.emit(TypedEvent.nextEventID(), request);
 
 		try {
-			const result = await this.workerPool.bulkSimAsync(request, onProgress);
+			const actionName = 'bulkSimAsync';
+			this.cancelCurrentSim(actionName);
+			const signal = this.addSimAbortSignal(actionName);
+			const result = await this.workerPool.bulkSimAsync(request, onProgress, { signal });
 			if (result.errorResult != '') {
 				throw new SimError(result.errorResult);
 			}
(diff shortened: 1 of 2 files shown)


FelixPflaum:
One binary can run multiple sims at the same time, so to stop them from outside we could add a unique id to each sim, and include that in the ProgressMetrics message. An exposed API could then stop sim loops by their id, similar to how I cancel them for threads in the concurrency code.

1337LutZ:
> One binary can run multiple sims at the same time, so to stop them from outside we could add a unique id to each sim, and include that in the ProgressMetrics message. An exposed API could then stop sim loops by their id, similar to how I cancel them for threads in the concurrency code.

That sounds like it could work. When you have time would you want to add that API so I could implement it?

FelixPflaum:
My first comment turned out to be not entirely correct. While in theory one wasm instance can do multiple requests, the async endpoints aren't truly async atm. Additionally the worker event loop is blocked while a sim is running because runtime.Goshed() does not give the event loop a chance to run, only go routines. While working on wasm concurrency I explored it a bit.

To implement a way to abort a running sim would require to
- make the ...Async endpoints actually async and return immediately, waiting for the final result in the WorkerPool handler instead.
- make the sim loop sleep periodically when running in wasm to give the event loop a chance to process a request to the worker.

I tested it and it works when doing time.Sleep(time.Microsecond) instead of runtime.Goshed() after wrapping the progress reporter into a go routine and returning directly on raidSimAsync. Though, all endpoints would need to be converted like that. Using sleep can be made conditional and enabled via the wasm main.go.

In that context it's then possible to add a simId to RaidSimRequest (and bulk or stat weight requests), which would be used to create a global entry with a channel to cancel sims with. A simple request could trigger that, similar to how the native concurrency code does it on errors for threaded sim loops already.

The JS side would just need to attach unique ids to requests and send an abort request to all workers. Actually keeping track of which worker does something wouldn't even be required at all.

1337LutZ:
@FelixPflaum Thanks for your thorough investigation!
We already have a sim unique ID right? So in theory that should work.
I agree that the asyncProgress check should actually be placed inside the worker_pool and not inside the worker itself, since it also doesn't cancel itself if it ever fails etc.

1337LutZ:
Closed since #463 is almost ready

PR #386, opened 2024-05-09
[DX] Typed Web Workers

1337LutZ:
Both WASM and Local work great now that you have live changes when running `WATCH=1 make webworkers`

- Converted the workers to Typescript
- Added types to the best of my ability
- Reused shared logic
- Added separate vite build bundling task for compiling each worker in a separate "library
- Added `make webworkers` task to compile, copy & watch the workers


kayla-glick:
After running `make wasm` and changing
image
the sim fails to start and throws these errors
image

1337LutZ:
> After running `make wasm` and changing image the sim fails to start and throws these errors image

Fixed it. `wasmready` was being tree shaken since it was unused. Added it to `globalThis` scope and now it works.

FelixPflaum:
I think if we're doing this, then why not refactor them entirely while we're at it. I'm especially offended by the net_ and local_ version sharing like 99.9% of their code :D

Made a quick PR for that against this branch https://github.com/wowsims/cata/pull/482

Advantages:
- All UI interface logic is contained in worker_interface.ts. That also means that adding or removing SimRequest defs in types.ts forces you to add the respective handlers for both worker versions.
- All wasm related code is entirely in sim_worker.ts
- All http related code is in worker_http.ts, the 2 versions then just set a different base URL.

1337LutZ:
Awesome, thanks! I will look into merging this 👍

ui/worker/worker_interface.ts
```diff
@@ -0,0 +1,62 @@
...
+	get workerId() {
+		return this._workerId;
+	}
+
+	/** Tell UI that the worker is ready. */
+	ready() {
+		if (!this.handlers) {
+			console.error('WorkerInterface.ready() used but handlers not set!');
+			return;
+		}
```
FelixPflaum, line 59: Forgot to remove this check. With handlers being set in the constructor it's not needed anymore.

FelixPflaum:
For all I care this can be merged. Seems to work fine.

1337LutZ:
@rosenrusinov if you validated that the local/sim workers work feel free to merg on your next round of checks :)

PR #397, opened 2024-05-10
Fixed bug with cat mangle while bearweaving

Timothy-Page:
Sim would do extra cat mangles when you have CC to try and optimize them but there are not needed if you are bearweaving and you can shred instead.


NerdEgghead (changes requested):
This makes sense, but you'll need to run `make test`, followed by `make update-tests`, and then commit the changed test file as well. I can see that only the `NoBuffs` tests changed, which is the expected behavior for this adjustment.

Timothy-Page:
@NerdEgghead Tests have been updated

PR #405, opened 2024-05-10
Warlock fixes

Tharre:


sim/warlock/pets.go
```diff
@@ -62,6 +68,16 @@ func (warlock *Warlock) makePet(summonType proto.WarlockOptions_Summon, baseStat
...
+	if warlock.Spec == proto.Spec_SpecDemonologyWarlock {
+		masteryBonus := func(mastery float64) float64 {
+			return 1 + math.Floor(2.3*core.MasteryRatingToMasteryPoints(mastery))/100.0
+		}
+
+		pet.AddOnMasteryStatChanged(func(sim *core.Simulation, oldMastery float64, newMastery float64) {
+			pet.PseudoStats.DamageDealtMultiplier /= masteryBonus(oldMastery)
+			pet.PseudoStats.DamageDealtMultiplier *= masteryBonus(newMastery)
+		})
+	}
```
rosenrusinov, line 80: This should be done in the pets Initialize method. Also right now you're removing the oldMastery bonus from the pet but I dont see it ever being activated before this callback. What will happen is that you are probably removing a non applied multiplier.

sim/warlock/talents_demonology.go
```diff
@@ -67,20 +67,21 @@ func (warlock *Warlock) registerManaFeed() {
...
-					warlock.AddMana(sim, restore, manaMetrics)
-				}
-			},
-		}))
+	aura := core.Aura{
+		Label:    "Mana Feed",
+		ActionID: actionID,
+		OnSpellHitDealt: func(aura *core.Aura, sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
+			if result.DidCrit() && spell.ClassSpellMask&WarlockBasicPetSpells > 0 {
+				warlock.AddMana(sim, manaReturn*warlock.GetStat(stats.Mana), manaMetrics)
```
rosenrusinov, line 75: `manaReturn*warlock.MaxMana()`

ui/index.html
```diff
@@ -428,7 +428,7 @@ <h3 class="expansion-title w-100">Cataclysm</h2>
 										<img src="https://wow.zamimg.com/images/wow/icons/large/class_warlock.jpg" class="sim-link-icon border-warlock">
 										<div class="d-flex flex-column">
 											<span class="sim-link-title">Warlock</span>
-											<span class="launch-status-label text-brand">Not Yet Supported</span>
+											<span class="launch-status-label text-brand">Alpha</span>
```
rosenrusinov, line 431: You also need to update `launched_sims.ts`

sim/warlock/talents_demonology.go
Tharre: Nice catch

ui/index.html
Tharre: Yeah

sim/warlock/pets.go
Tharre: I guess moving it into initialize is cleaner, but should be functionally identical right?

I'm indeed removing the old mastery bonus, but since mastery at the start is 0 that multiplier is 1.0, so it doesn't do anything.

rosenrusinov: This callback is called only for dynamic stat changes during runs. If you apply the DamageMulti in the Initialize of the pet it should match the value of oldMastery in the first callback

Tharre: Pets can't possibly have mastery at the start, so I don't need anything in initialize. The only way they can get it is via stat inheritance and that is dynamically applied on summon.

PR #418, opened 2024-05-11
Balance druid implementation

stn-ms:
- set the sim to alpha, I believe it's time
- added the 25% baseline dmg boost to eclipses
- hurricane update
- starfall update
- wild mushrooms implemented
- force of nature - only uncommented the old implementation for completeness right now, I realize it needs some work especially on the damage values of treants, but noone knows the exact values, thus I'm planning on doing some tests on the beta soon
- moved more glyphs into glyphs.go - this is WIP :)
- implemented Shooting Stars
- somewhat implemented Euphoria - the mana back on eclipse is working (that's why I moved the bar to ApplyTalents() for now), then I made a ProcTriggerAura for the double energy part and I was planning on checking whether aura is active in Wrath and Starfire - I'm not sure if this is the best approach or just sim.Proc in the spells themselves. Feedback would be appreciated.


sim/druid/talents.go
```diff
@@ -191,6 +194,106 @@ func (druid *Druid) applyGenesis() {
...
+	if druid.HasEclipseBar() {
+		druid.AddEclipseCallback(func(_ Eclipse, gained bool, sim *core.Simulation) {
+			if gained {
+				druid.AddMana(sim, druid.MaxMana()*euphoriaManaGainPct, manaMetrics)
+			}
+		})
+	}
+
+	// Double eclipse energy
+	euphoriaDummyAura := druid.GetOrRegisterAura(core.Aura{
```
rosenrusinov, line 220: This aura doesnt do anything as far as I can see but is activated before the Wrath and Starfire apply effects are called and is deactivated on a cast complete which will come from the same spell that is activating the aura from the proc trigger. Dont know if that is what you intended or this is just wip stuff

PR #428, opened 2024-05-12
Warlock fixes

Tharre:


sim/warlock/pets.go
```diff
@@ -339,9 +351,14 @@ func (pet *WarlockPet) registerFireboltSpell() {
...
 		Cast: core.CastConfig{
+			CD: core.Cooldown{
+				// emulate the imp 1.5s GCD that does not change with haste
+				Timer:    pet.NewTimer(),
+				Duration: 1500 * time.Millisecond,
+			},
 			DefaultCast: core.Cast{
 				GCD:      core.GCDDefault,
-				CastTime: time.Millisecond * 2500,
+				CastTime: 2500 * time.Millisecond,
```
rosenrusinov, line 361: Instead of having this artificial cooldown you can make the cast IgnoreHaste: true and add a ModifyCast handler that manually applies the cast time to the cast. You can see how DKs Gargoyle has its Spell configured its exactly the same concept (spell is hasted but GCD cant get below 1.5)

Tharre: I guess I can change it for consistency but seems to achieve pretty much the same.

Tharre: Actually no; this is broken :joy:
The CD obviously only applies after the cast is finished so it's waiting 1.5s between every cast

PR #433, opened 2024-05-12
[UI] Replace SASS vars with CSS Variables

1337LutZ:
- Replaced 99.999% of all `$sass-var` with `var(--css-var)` for some future proofing and ease in managing overrides
- Replaced stray colors that were very similar with colors that are defined/exist


kayla-glick (approved):
Everything looks normal as far as I can tell 👍 thanks for this!

PR #434, opened 2024-05-12
Adds basic movement encounter

ToxicKevinFerm:
Moves 10 yards between 5-40 yards every 10 seconds.
Not sure its relevant for melees 🤔


sim/encounters/light_movement.go
```diff
@@ -0,0 +1,83 @@
...
+func (ai *LightMovementAI) Reset(sim *core.Simulation) {
+	ai.LastMoveTime = 0
+}
+
+func (ai *LightMovementAI) ExecuteCustomRotation(sim *core.Simulation) {
+	players := sim.Raid.AllPlayerUnits
+
+	if !ai.ShouldMove(sim) {
+		return
+	}
```
rosenrusinov, line 68: Instead of polling this in every rotation callback you can delay the rotation callback with a ai.Target.WaitUntil()

rosenrusinov (approved):

PR #453, opened 2024-05-13
[Warrior] T12 4P

MatthewKyle:
Added Prot T12 4P
Updated Tests

### DIFF
--- a/sim/warrior/items.go
+++ b/sim/warrior/items.go
@@ -109,14 +109,18 @@ var ItemSetMoltenGiantWarplate = core.NewItemSet(core.ItemSet{
 			actionID := core.ActionID{SpellID: 99237}
 
 			fieryAttack := character.RegisterSpell(core.SpellConfig{
-				ActionID:       actionID,
-				SpellSchool:    core.SpellSchoolFire,
-				ProcMask:       core.ProcMaskEmpty, // TODO (4.2) Test this
-				Flags:          core.SpellFlagMeleeMetrics,
-				CritMultiplier: character.DefaultMeleeCritMultiplier(),
+				ActionID:    actionID,
+				SpellSchool: core.SpellSchoolFire,
+				ProcMask:    core.ProcMaskEmpty, // TODO (4.2) Test this
+				Flags:       core.SpellFlagMeleeMetrics,
+
+				CritMultiplier:   character.DefaultMeleeCritMultiplier(),
+				DamageMultiplier: 1,
+				ThreatMultiplier: 1,
+
 				ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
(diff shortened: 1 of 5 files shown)


sim/warrior/items.go
```diff
@@ -134,56 +134,56 @@ var ItemSetMoltenGiantWarplate = core.NewItemSet(core.ItemSet{
...
-// 		4: func(agent core.Agent) {
-// 			panic("Not yet implemented pending a way to model 'trigger aura on expiration of another'")
-// 		},
-// 	},
-// })
+			core.MakeProcTriggerAura(&character.Unit, core.ProcTrigger{
+				Name:           "Combust Trigger",
+				ActionID:       actionID,
+				Callback:       core.CallbackOnSpellHitDealt,
+				ClassSpellMask: SpellMaskShieldSlam,
```
rosenrusinov, line 175: should this also have an Outcome filter for Landed?

sim/warrior/shield_block.go
```diff
@@ -10,9 +10,24 @@ import (
...
+		Label:    "T12 4P Bonus",
+		ActionID: core.ActionID{SpellID: 99242},
+		Duration: 10 * time.Second,
+		OnGain: func(aura *core.Aura, sim *core.Simulation) {
+			warrior.AddStatDynamic(sim, stats.Parry, 6*core.ParryRatingPerParryChance)
+		},
+		OnExpire: func(aura *core.Aura, sim *core.Simulation) {
+			warrior.AddStatDynamic(sim, stats.Parry, -6*core.ParryRatingPerParryChance)
+		},
+	})
```
rosenrusinov, line 23: You should put this inside an if hasT12_4P_bonus so it doesnt register the aura without the set equipped. If you do this you dont need to check if hasT12 is true in the proc trigger because our aura.Activate() already does a null check for the aura you try to activate.

MatthewKyle: Okay, so I added aura to Warrior.  If I understand you correctly I can involve it in Sheildblock expires just as a warrior.T12_4P_Bonus.activate(sim)  with the sim. duration checks correct?

sim/warrior/items.go
MatthewKyle: Added.  I assume it requires a landed hit to apply yes.

sim/warrior/shield_block.go
rosenrusinov: You dont need a new reference in warrior to this aura. Keep it a local variable in the register function and only assign it if the set is equipped. Then in the expire handler you only need to check if its end of iteration and if not activate the aura. Even if the set is not equipped it will still work with a null aura, it just wont try to activate

MatthewKyle: Okay, I think I did what you wanted.  Let me know if I missed the mark.

PR #455, opened 2024-05-13
Some shaman fixes and character stat fix

Polynomix:
When building the stats breakdown for dispaly in the UI, some dependencies were not being applied to the entire stats, so when removing it, it would result in stats being higher/lower than what they really are.

Fix base shaman attack power calculations
Fix fire ele melee damage
Fix typo in the fortitude buff
Add talent requirement to totemic wrath raid buff


sim/core/unit.go
```diff
@@ -310,7 +310,9 @@ func (unit *Unit) AddStatsDynamic(sim *Simulation, bonus stats.Stats) {
 
 	unit.statsWithoutDeps.AddInplace(&bonus)
 
-	bonus = unit.ApplyStatDependencies(bonus)
+	if !unit.Env.MeasuringStats || unit.Env.State == Finalized {
+		bonus = unit.ApplyStatDependencies(bonus)
+	}
```
rosenrusinov, line 315: so stat dependencies should not be applied during stat measurements? That sounds wrong to me

Polynomix: it is applied [here](https://github.com/wowsims/cata/blob/4fccc54550269f7a6f8cbbd3e80346ef962a90ef/sim/core/character.go#L255), the issue is that this test is present for multiplicative modifiers but not additive ones,  so for example with UDK [here](https://wowsims.github.io/cata/death_knight/unholy/#eJzjYhFgdGB0YhBikLrOPIuZLSAnsTK1iINRgEtKiYuBEFRi0NrLxNXBxDAKwEDIgFQdVgxODFH8UrxcnIIQgQ/2QgyHmFQ43jxlEph/hUli7hUmhTlXmDQ2XWMy6L3CZDHzKpPD9KtMHjOuMJ1gYrnAxH6LieERE5ME4wumlEZmiBkiDguhrEiHLmYGpTtMnGBeQ5qDIEyF5KyZIHDS3hIicsFeMQ0Mrtk7nj0DAm/sjRYzKURIsGrROII1Oh3I0daQcpwsfZQASGDtfe/gCOH/sE9gBAD6EjyZ) with a naked, talentless character we should get (188+549)*1.25*1.05 = 967,3 STR, but instead we have ((188+(549\*1.05\*1.25)-(549\*1.25))+549)\*1.25\*1.05 = 1012.3 STR. This is because multiplicative dependencies are not applied to base stats because of [that test](https://github.com/wowsims/cata/blob/4fccc54550269f7a6f8cbbd3e80346ef962a90ef/sim/core/buffs.go#L32) which is the same i used.

when clearing auras [here](https://github.com/wowsims/cata/blob/4fccc54550269f7a6f8cbbd3e80346ef962a90ef/sim/core/character.go#L290), some multiplicative dependencies are deactivated before the flat stats one are so it does not remove the entire added stats, and at the end, we end up with more strength than we started with, (222.3 instead of 188 for this example)

With the test i added, flat buff are added and removed to the base stats without applying any dependencies, the measuring function will do it.

rosenrusinov (approved):

rosenrusinov:
can merge with solved conflicts

sim/shaman/fire_elemental_pet.go
```diff
@@ -32,17 +32,21 @@ func (shaman *Shaman) NewFireElemental(bonusSpellPower float64) *FireElemental {
...
-			BaseDamageMin:  429, //Estimated from beta testing
-			BaseDamageMax:  463, //Estimated from beta testing
-			SwingSpeed:     2,
-			CritMultiplier: 2.66, //Estimated from beta testing
-			SpellSchool:    core.SpellSchoolFire,
+			BaseDamageMin:     429, //Estimated from beta testing
+			BaseDamageMax:     463, //Estimated from beta testing
+			SwingSpeed:        2,
+			CritMultiplier:    2.66, //Estimated from beta testing
+			AttackPowerPerDPS: 2,
```
raethkcj, line 44: Do you have a reason to think AttackPowerPerDPS isn't 14? If not I'd rather see it left at default and scale inherited AP up by 7, since AP/14 is standard across all other player-controlled things

Polynomix: i can do that

PR #457, opened 2024-05-14
Several improvements to item picker

kayla-glick:
- Updated reforge text to include values (Screenshot 1)
- Added ilvl to the gear picker modal and made the modal larger to accommodate space a bit better (Screenshot 2)
- Added sorting by both ILvl and EP in the modal (Screenshot 2)
- Added min and max ilvl filters (Screenshot 3)
- By default disabled all armor types except the spec's primary armor type for Armor Specializationbonus

image
image
image

### DIFF
--- a/ui/core/components/filters_menu.tsx
+++ b/ui/core/components/filters_menu.tsx
@@ -1,10 +1,13 @@
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import { element } from 'tsx-vanilla';
+
 import { Player } from '../player.js';
 import { ItemSlot } from '../proto/common.js';
 import { RaidFilterOption, SourceFilterOption, UIItem_FactionRestriction } from '../proto/ui.js';
 import { armorTypeNames, raidNames, rangedWeaponTypeNames, sourceNames, weaponTypeNames } from '../proto_utils/names.js';
 import { Sim } from '../sim.js';
 import { EventID } from '../typed_event.js';
-import { BaseModal } from './base_modal.js';
+import { BaseModal } from './base_modal.jsx';
 import { BooleanPicker } from './boolean_picker.js';
 import { EnumPicker } from './enum_picker.js';
 import { NumberPicker } from './number_picker.js';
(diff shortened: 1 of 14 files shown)


ui/core/components/filters_menu.tsx
```diff
@@ -1,10 +1,13 @@
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import { element } from 'tsx-vanilla';
+
 import { Player } from '../player.js';
 import { ItemSlot } from '../proto/common.js';
 import { RaidFilterOption, SourceFilterOption, UIItem_FactionRestriction } from '../proto/ui.js';
 import { armorTypeNames, raidNames, rangedWeaponTypeNames, sourceNames, weaponTypeNames } from '../proto_utils/names.js';
 import { Sim } from '../sim.js';
 import { EventID } from '../typed_event.js';
-import { BaseModal } from './base_modal.js';
```
1337LutZ, line 7: Actually all these imports can be done without extensions, I've tried to remove them in files I touch

ui/core/components/filters_menu.tsx
```diff
@@ -19,7 +22,36 @@ export class FiltersMenu extends BaseModal {
 	constructor(rootElem: HTMLElement, player: Player<any>, slot: ItemSlot) {
 		super(rootElem, 'filters-menu', { size: 'md', title: 'Filters', disposeOnClose: false });
 
-		let section = this.newSection('Factions');
+		let section = this.newSection('General');
```
1337LutZ, line 25: This can be a const

ui/core/components/filters_menu.tsx
```diff
@@ -19,7 +22,36 @@ export class FiltersMenu extends BaseModal {
...
+
+		const ilvlFiltersContainer = <div className="ilvl-filters" />;
+		section.appendChild(ilvlFiltersContainer);
+
+		new NumberPicker(ilvlFiltersContainer as HTMLElement, player.sim, {
```
1337LutZ, line 30: Instead of casting it 2x here and on line 44, I would just cast it here:
```const ilvlFiltersContainer = <div className="ilvl-filters" /> as HTMLElement;```

File: ui/core/components/number_picker.ts [hunk 4]
```diff
@@ -1,31 +1,37 @@
...
+		this.positive = config.positive ?? false;
+		this.showZeroes = config.showZeroes ?? true;
 
 		this.inputElem = document.createElement('input');
 		this.inputElem.type = 'text';
 		this.inputElem.classList.add('form-control', 'number-picker-input');
 
 		if (this.positive) {
-			this.inputElem.onchange = e => {
+			this.inputElem.onchange = _event => {
```
1337LutZ, line 34: The unused events can just be removed, also on L46 and L50

ui/core/sim.ts
```diff
@@ -624,7 +624,12 @@ export class Sim {
 
 			const filters = proto.filters || Sim.defaultFilters();
 			if (filters.armorTypes.length == 0) {
-				filters.armorTypes = Sim.ALL_ARMOR_TYPES.slice();
+				// TODO: Can we specify the time of sim so that we can do this without querying the DOM?
+				if (!!document.querySelector('.not-within-raid-sim')) {
```
1337LutZ, line 628: Sadly not, because we don't have access to simUI here

1337LutZ: We could add it as a property when initializing and pass it when we initialize `raid_sim_ui`

rosenrusinov: I would love for us to have a proper IsWithinRaidSim utility function in both the FE and BE code ^^

ui/core/components/filters_menu.tsx
1337LutZ: Ahhh sorry, I didn't see it in the small snipped

1337LutZ: Oh, is it abusing 1 let to init all the tabs? :O

PR #460, opened 2024-05-14
[UI] Various responsive & input fixes

1337LutZ:
- Added ID and for to inputs so the label can also be clicked to focus inputs or check checkboxes
- Replaced "outdated" margin-left and bottom magic on a lot of components and instead use CSS Grid and `gap` to make it more flexible (fixed quite some odd leftover margins)
- Optimised player-picker inside the raid sim (removed the huge sim sidebar height for example)
- Fixed list picker minimum items logic
- Fixed encounter picker overflow scroll

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,8 @@
                 "bootstrap": "^5.3.0",
                 "pako": "^2.0.4",
                 "tippy.js": "^6.3.7",
-                "tsx-vanilla": "^1.0.0"
+                "tsx-vanilla": "^1.0.0",
+                "uuid": "^9.0.1"
             },
             "devDependencies": {
                 "@protobuf-ts/plugin": "2.9.1",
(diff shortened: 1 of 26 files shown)


kayla-glick:
Looks like there's something weird happening in the encounter picker. It's overflowing on the x axis. I'm also not sure what this field is as i've never seen it before
image

kayla-glick:
Shaman totem spacing is a bit odd now as well
image

kayla-glick:
Looks like glyph area width is off too. These were aligned with the talents before
before
after
image
image

1337LutZ:
> Looks like there's something weird happening in the encounter picker. It's overflowing on the x axis. I'm also not sure what this field is as i've never seen it before. I think that's more of an issue of just needing a shorter name on that field tbh. image

@kayla-glick There's actually multiple fields that overflow due to `white-space: nowrap`. I see 2 options:
1. Remove the nowrap, which can cause layout shifts in the form (or I could align them to the bottom within the input wrapper to prevent that.
2. Ellipsis the overflow into `...` and have it cut off when it will overflow.

1337LutZ:
> Looks like there's something weird happening in the encounter picker. It's overflowing on the x axis. I'm also not sure what this field is as i've never seen it before. I think that's more of an issue of just needing a shorter name on that field tbh. image

Fixed

1337LutZ:
> 

Fixed

1337LutZ:
> Shaman totem spacing is a bit odd now as well. I think it's just an issue with icon groups image image

Fixed

1337LutZ:
> Looks like glyph area width is off too. These were aligned with the talents before before after image image

Fixed

PR #463, opened 2024-05-14
Wasm concurrency and sim abort functionality

FelixPflaum:
Brings concurrency to the Wasm version by reusing the Go code for splitting requests and combining results. This way the JS side only manages progress and delegates workers. The prone to error work is done by already existing Go code that is backed by tests. Also adds an abort API and abort functionality to all sims.

### Changes
- Add API that exposes the logic for raid sim request splitting and result combination to Wasm workers.
- Add global concurrency setting to the settings menu. Only shown when Wasm is used. Dropdown with "Off" to navigator.hardwareConcurrency as options. Workers will be spawned or destroyed when the setting changes. Termination is delayed until tasks are done. SimWorker objects are reused to allow changing the worker amount even while tasks are running.
- Update WorkerPool and Workers to know when they are running the Wasm version. 
- Overhaul worker task tracking logic for load balancing. It now tracks "work amount left" using (remaining) iterations of requests.
- Wasm concurrency for raid sims using split requests on multiple workers.
- Refactor stat weight sim, exposing a request generation step and weight calculation step as api endpoints. Concurrency is now achieved by using concurrent raid sims. 
- Wasm concurrency for stat weight sim by reusing concurrent raid sim wasm logic.
- Add signal API to both JS and Go to enable abort functionality. Async requests can be aborted using the request Id that is now passed with the request.
- Starting an async sim aborts all running async sims.
- Add stop button to progress views.
- Disable "Simulate" button while a raid sim is running.

### Things to note
- **Firefox**: Every worker uses ~250 MB RAM for me. Going up to the logical number of cores (16) can actually cause my browser to stop workers due to out of memory errors. There's a note below the concurrency setting that mentions this. Initializing workers also causes a significant CPU usage for some time. Neither seems to happen in chromium based browsers.
- Bulk sim is not changed to be concurrent in wasm. Abort works though.


ui/core/sim_concurrent.ts
```diff
@@ -0,0 +1,167 @@
...
+	for (let i = 0; i < requests.length; i++) {
+		wp.raidSimAsync(requests[i], pm => { progressHandler(i, pm) });
+	}
+
+	return new Promise(res => {
+		resolve = res;
+	});
+}
+
+export async function runConcurrentSim(request: RaidSimRequest, workerPool: WorkerPool, onProgress: WorkerProgressCallback): Promise<RaidSimResult> {
```
1337LutZ, line 143: No need to explicetly define the return type, as it is inferred by default

1337LutZ, line 143: `export const runConcurrentSim = async (...)=>`

ui/core/components/settings_menu.tsx
```diff
@@ -152,5 +160,31 @@ export class SettingsMenu extends BaseModal {
 					sim.setShowQuickSwap(eventID, newValue);
 				},
 			});
+
+		if (useConcurrentWorkersWrap.value && useConcurrentWorkers.value) {
+			const values: EnumValueConfig[] = [{value: 0, name: "Off"}];
+			for (let i = 2; i < navigator.hardwareConcurrency + 1; i++) {
```
1337LutZ, line 166: This logic can be reduced and returned into a array like this:
```
const values = navigator.hardwareConcurrency.reduce<EnumValueConfig[]>(
  (acc, _, i) => {
    acc.push({ value: i, name: i.toString() });
    return acc;
  },
  [{ value: 0, name: "Off" }]
);
```

1337LutZ:
I can probably improve the Vite server to add caching to specific content.

1337LutZ:
The memory issue in FF might be related to the following:
- https://bugzilla.mozilla.org/show_bug.cgi?id=1407691
- https://stackoverflow.com/a/35106711

Seems that FF has GC issues when creating new ArrayBuffers in a Worker.

ui/core/components/settings_menu.tsx
FelixPflaum: navigator.hardwareConcurrency is a number value. How is that supposed to work?
Also I explicitly omit 1 because it doesn't make sense in this context, so a good old boomer loop seems to be the easiest solution :D

1337LutZ: Ah yea, should've added:  `Array(navigator.hardwareConcurrency)` I blindly wrote it

FelixPflaum:
> I can probably improve the Vite server to add caching to specific content.

@1337LutZ Not a vite problem. The go http server sets a no-cache header and has no gzip support, so the wasm lib can't be cached either way due to its size. This doesn't really affect local dev environments or the hosted version. 

I only encountered it as a problem because I use a VM on a remote server to work on this, but I use my local browser to access the UI (more responsive). With the net worker that's fine, but using wasm my download speed becomes a massive bottleneck on each site load. I'm simply using the browser on the VM now when dealing with wasm.

1337LutZ:
@FelixPflaum Ah yes sorry, I totally forgot to reply here. When working on the Typed workers and the components repo connection I found out the same thing when going through the code

ui/core/sim_concurrent.ts
FelixPflaum: Still doesn't hurt to define them. Prevents wrong or missing returns and shows the error for the function, instead of down the line where the function is used.

FelixPflaum: What's the advantage of exporting it as an arrow function when there's no this context being used?

1337LutZ: If you don't use `this` (or any other binding) it's a common (if not the) frontend convention to use arrow functions, as they don't have any bindings and are scoped (`this`, `prototype`, `constructor`, `new`, `arguments`). They should be the default, not the expection

ui/core/components/results_viewer.tsx
```diff
@@ -37,19 +39,22 @@ export class ResultsViewer extends Component {
...
 		const warningElemRef = ref<HTMLDivElement>();
+		const buttonElemRef = ref<HTMLDivElement>();
 
 		this.rootElem.appendChild(
 			<>
 				<div ref={pendingElemRef} className="results-pending">
 					<div className="loader"></div>
 				</div>
 				<div ref={contentElemRef} className="results-content"></div>
+				<div ref={buttonElemRef} className="button-zone" style="text-align: center"></div>
```
1337LutZ, line 50: There's a classname `text-center` (can you also change it for the warning-zone?

ui/core/components/results_viewer.tsx
```diff
@@ -127,4 +133,20 @@ export class ResultsViewer extends Component {
 		this.contentElem.style.display = 'block';
 		this.pendingElem.style.display = 'none';
 	}
+
+	addAbortButton(abortClicked: (event: MouseEvent) => void) {
+		this.buttonElem.innerHTML = '';
+		this.buttonElem.style.display = 'block';
+		this.buttonElem.appendChild(
```
1337LutZ, line 140: Remove `this.buttonElem.innerHTML = '';` and then change this to `this.buttonElem.replaceChildren()`. Actually the more "performant" due to only triggering 1 render.

ui/core/components/results_viewer.tsx
```diff
@@ -127,4 +133,20 @@ export class ResultsViewer extends Component {
...
+			<button className="sim-abort-button" onclick={abortClicked}>
+				<i className="fa fa-times fa-lg me-1" />
+				Stop
+			</button>,
+		);
+	}
+
+	removeAbortButton() {
+		this.buttonElem.innerHTML = '';
```
1337LutZ, line 149: This can be  `this.buttonElemen.replaceChildren()`, We are trying to get rid of all `innerHTML` things

ui/core/sim.ts
```diff
@@ -271,8 +279,13 @@ export class Sim {
 		this.bulkSimStartEmitter.emit(TypedEvent.nextEventID(), request);
 
 		try {
-			const result = await this.workerPool.bulkSimAsync(request, onProgress);
+			const signals = this.signalManager.registerRunning(request.requestId, RequestTypes.BulkSim);
+			const result = await this.workerPool.bulkSimAsync(request, onProgress, signals);
 			if (result.errorResult != '') {
+				if (result.errorResult.includes('aborted')) {
+					// TODO: Abort feedback?
```
1337LutZ, line 286: Could perhaps add a toast if you want to show feedback

ui/core/worker_pool.ts
```diff
@@ -155,24 +154,65 @@ export class WorkerPool {
...
+			worker.doApiCall(requestName, request, id);
+			const finalProgress: Promise<ProgressMetrics> = new Promise(resolve => {
+				// Add handler for the progress events
+				worker.addPromiseFunc(
+					this.getProgressName(id),
+					this.newProgressHandler(id, worker, onProgress, pm => resolve(pm)),
+					noop,
+				);
+			});
+			return await finalProgress;
```
1337LutZ, line 186: Is there a need to await this if you return an async/Promise function anyways?

ui/core/components/results_viewer.tsx
FelixPflaum: Added the class for both.

ui/core/components/results_viewer.tsx
FelixPflaum: Replaced it.

ui/core/sim.ts
FelixPflaum: I changed error_result into a ErrorOutcome object to disambiguate the abort outcome. Added a toast if it was an abort.

ui/core/worker_pool.ts
FelixPflaum: Yes, the finally block needs to run after the request is done.

ui/core/sim.ts
1337LutZ: Nice!

FelixPflaum:
Couldn't find time and motivation to make bulk sim concurrent for wasm. Otherwise this should work fine now. Firefox still has the RAM and (temporary) CPU usage problem, but nothing I can do about that.

Should probably be checked by others before merging, there's enough changes for me to miss some errors, especially after the last master merge.

ui/core/sim_concurrent.ts
```diff
@@ -0,0 +1,295 @@
...
+	return errRes;
+};
+
+export const runConcurrentSim = async (
+	request: RaidSimRequest,
+	workerPool: WorkerPool,
+	onProgress: WorkerProgressCallback,
+	signals: SimSignals,
+): Promise<RaidSimResult> => {
+	console.log(`Sending requests split for ${workerPool.getNumWorkers()} splits.`);
```
1337LutZ, line 158: Could you add `isDevMode()` for all the normal logs in this file?

sim/core/statweight.go
```diff
@@ -116,115 +114,40 @@ func CalcStatWeight(swr *proto.StatWeightsRequest, referenceStat stats.Stat, pro
...
-	}
-
 	const defaultStatMod = 40.0 // match to the impact of a single gem
 	statModsLow := make([]float64, stats.UnitStatsLen)
 	statModsHigh := make([]float64, stats.UnitStatsLen)
 
 	// Make sure reference stat is included.
-	statModsLow[referenceStat] = defaultStatMod
-	statModsHigh[referenceStat] = defaultStatMod
+	statModsLow[swr.EpReferenceStat] = -defaultStatMod
```
1337LutZ, line 149: Is this intended? The old code didn't have a negative value here

1337LutZ:
2 comments, rest looks and works real nice (tested Local and WASM)! Stat weights sim so much quicker as well.

sim/core/statweight.go
FelixPflaum: I'm pretty sure that was a bug in the old code. The purpose of the low and high is to sim with a positive and negative mod. It probably never mattered because the reference stat is always included and added with a negative mod for the low case like all the other stats to sim.

1337LutZ: Yea I figured, but was just double checking 👍

rosenrusinov:
@1337LutZ is this good to be merged now?

1337LutZ:
@rosenrusinov Yep, we can always monitor and revert it if we hear about very big issues.

PR #479, opened 2024-05-16
updating orc enh shaman p1 bis

foreverpsycotic:


rosenrusinov (commented):
Default gear should not be race specific.

foreverpsycotic:
> Default gear should not be race specific.

Its because of the racials.  Draenei get more hit by default, orc gets more expertise so there is a shift in gems/reforging depending on phase.

PR #480, opened 2024-05-16
Fix database modification of item stats object

1337LutZ:
This fixes the modification of the `UIItem` core stats resulting in double random affix / reforge values.

### DIFF
--- a/ui/core/player.ts
+++ b/ui/core/player.ts
@@ -446,9 +446,10 @@ export class Player<SpecType extends Spec> {
 	}
 
 	// Returns all reforgings that are valid with a given item
-	getAvailableReforgings(item: Item): Array<ReforgeData> {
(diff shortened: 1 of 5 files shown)


ui/core/player.ts
```diff
@@ -457,10 +458,11 @@ export class Player<SpecType extends Spec> {
 		return this.sim.db.getReforgeById(id);
 	}
 
-	getReforgeData(item: Item, reforge: ReforgeStat): ReforgeData {
+	getReforgeData(equippedItem: EquippedItem, reforge: ReforgeStat): ReforgeData {
+		const withRandomSuffixStats = equippedItem.getWithRandomSuffixStats();
+		const item = withRandomSuffixStats.item;
```
1337LutZ, line 463: This creates a defensive clone

PR #503, opened 2024-05-20
Added support for not showing auras in APLs

Polynomix:
Might be counter intuitive with how the flag works for spells since here you need the flag to not having it displayed in the dropdown menu.


Tharre:
I wouldn't do this with flag masks, just add a boolean ShowInAPL at the end of the aura struct. Has the nice benefit of keeping the size of that struct the same, instead of adding yet another 4 bytes to it.

PR #504, opened 2024-05-21
[Mage] Various Fixes

gjoeyjoe:
Combustion fixed in today's patch notes to not double-dip mastery on ignite.

Impact will correctly duplicate LB, Combustion, and Pyrodot. Ignite is still TODO.

Blast Wave's Flamestrike now acts separate from normal Flamestrike.

Fixed DB glyph CD and Blast Wave CD.


sim/core/dot.go
```diff
@@ -383,3 +383,30 @@ func (spell *Spell) createDots(config DotConfig, isHot bool) {
 		}
 	}
 }
+
+func (dot *Dot) CopyDotAndApply(sim *Simulation, originaldot *Dot) {
+	dot.TakeSnapshot(sim, false)
+	dot.SnapshotBaseDamage = originaldot.SnapshotBaseDamage
+
+	dot.tickPeriod = originaldot.tickPeriod
+	dot.NumberOfTicks = originaldot.NumTicksRemaining(sim) // originaldot.NumberOfTicks
```
rosenrusinov, line 392: Changing `NumberOfTicks` is a destructive action. You should not change this but instead change the `TickCount` to match the count of the original dot

rosenrusinov (changes requested):

sim/core/dot.go
gjoeyjoe: Did some testing and it didn't seem necessary to even have this line. Removed for now.

rosenrusinov (approved):

PR #548, opened 2024-05-27
fix blood frenzy being overwritten by other bleed effects 

zoristaken:
fix blood frenzy being overwritten by other bleed effects and losing the phys vuln debuff part

image

### DIFF
--- a/sim/core/debuffs.go
+++ b/sim/core/debuffs.go
@@ -41,6 +41,7 @@ func applyDebuffEffects(target *Unit, targetIdx int, debuffs *proto.Debuffs, rai
 	// +4% Phsyical Damage
 	if debuffs.BloodFrenzy && targetIdx < 4 {
 		MakePermanent(BloodFrenzyAura(target, 2))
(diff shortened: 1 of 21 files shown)


rosenrusinov:
I think instead of adding some weird priority to this you want to just make the bleed effects not exclusive? This would still keep the bleed effects not stackable but the auras themself should be both present on the target. If in-game they actually overwrite each other and one gets removed when you apply the other then your implementation is correct. But I think it usually allows for both buffs to be on the target just not stacking their bonuses

rosenrusinov:
hmm I wasnt aware the bleed effect is not implemneted as an exclusive effect. It should be refactored to use the same concenpt then you can have multiple of them without the effect mod actually stacking

rosenrusinov:
If in-game they overwrite each other then we should leave them as they were with exclusive: true.

rosenrusinov:
I meant to just change is singleAura property to false. That would allow the exclusive effect to not stack but the aura to stack with other auras that have the same exclusive effect.

PR #549, opened 2024-05-27
[Balance] Enable mastery snapshotting + Innervate

koflem:
Ability to start in solar eclipse
Ability to specify a mastery amount for the initial solar eclipse
Enable self innervate (some stuff about innervate target didn't seem to be working so it's just innervate self for now, not like there's much of a reason to innervate someone else anymore...)

### DIFF
--- a/proto/druid.proto
+++ b/proto/druid.proto
@@ -138,6 +138,8 @@ message BalanceDruid {
   message Options {
     DruidOptions class_options = 1;
 	float okf_uptime = 2;
+	bool start_in_solar = 3;
+	int32 mastery_snapshot = 4;
   }
   Options options = 3;
 }
--- a/sim/core/buffs.go
+++ b/sim/core/buffs.go
@@ -1670,7 +1670,7 @@ func registerInnervateCD(agent Agent, numInnervates int32) {
 	character := agent.GetCharacter()
 	character.Env.RegisterPostFinalizeEffect(func() {
 		innervateThreshold = InnervateManaThreshold(character)
-		innervateAura = InnervateAura(character, -1, false)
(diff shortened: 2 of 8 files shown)


rosenrusinov (changes requested):
- Innervate should support setting external target for raid sim.
- This mastery snapshot is never cleared so every iteration you keep the snapshot mastery for the hole run

koflem:
> * Innervate should support setting external target for raid sim.
Okay, I've never used raid sim so I would have no idea how to test it. I can put back some of the old code and hope it works but given it didn't work before that seems dubious. I guess I'll check out how to raid sim...
>  * This mastery snapshot is never cleared so every iteration you keep the snapshot mastery for the hole run
That's not true, it is cleared.

koflem:
The raid sim isn't even enabled yet for cata... I added back some of the code that would allow a buff target so it *should* work but I have no idea if it does.
Well, it's not like it worked before anyways.

koflem:
It seems like there's an issue in raid sims where only the initial simulation works, any other simulation will break until you refresh. Knowing this, I was able to innervate a mage in raid sim mode so it should be good.

koflem:
Even on the deployed master raid sims I can't get anyone to deal damage (well the tank does like 1k) on any simulations after the first one after a refresh. But like I said I tested it and it did properly innervate the target in the raid sim. One issue is the current APL for balance only uses major cooldowns at 100 lunar energy so the innervate might not be timed properly but that's an APL issue.

rosenrusinov (approved):

stn-ms:
I would much rather have the option to start at whatever energy level, instead of just forcing Solar. If tomorrow someone figures out that there's merit to starting in Lunar, we would have to do that too.

PR #569, opened 2024-05-29
[Mage] Preset Gear/Talent Update

gjoeyjoe:
Updating presets to match wowhead's presets.
Changed APL to turn hot streak into a reaction time cast.


gjoeyjoe:
I keep forgetting tests are a thing 😩

PR #615, opened 2024-06-05
Handle item sets sharing bonuses using ItemSet in ItemSparse db

hillerstorm:
So I had a look at shared set bonuses yesterday (mainly relevant for PvP sets, since prebis for retpalas use 2pc ilvl 365).
We already parse the ItemSparse db from wago, and that table includes an ItemSet column containing the ID for the set actually giving bonuses, so I just extended the wago db import to also fetch those ids and save in our db.

If we use retpalas as an example and look at https://wago.tools/db2/ItemSparse?build=4.4.0.54901&filter[ItemSet]=917&page=1&sort[Display_lang]=asc all PvP set items for this expansion all share 917, which is the base set without a prefix (Gladiator's Vindication, that actually is unobtainable).
Doing this we can just register this base set without a prefix, add the ID and then all relevant items gets counted to the same bonus 🙂 no need to extend the "alternative name" logic. In fact the name could be anything but that's how blizz does it so why not stay consistent.
I tested with hunters and it works like a charm, having one piece from each season adds the 2pc and 4pc stats just like it should!

Live, no bonus:
image

With the change, bonuses work (+70 +90):
image


sim/hunter/wotlk_items.go
```diff
@@ -9,26 +9,26 @@ import (
...
+// 		},
+// 		4: func(agent core.Agent) {
+// 			hunter := agent.(HunterAgent).GetHunter()
+// 			hunter.AddStats(stats.Stats{
+// 				stats.AttackPower:       150,
+// 				stats.RangedAttackPower: 150,
+// 			})
+// 		},
+// 	},
+// })
```
rosenrusinov, line 31: Why did this get commented?

hillerstorm: For some reason wotlk_items.go is still active (for the T10 bonus perhaps?) and this name clashed with the new pvp-set in cata.
I could just remove it otherwise :)

rosenrusinov: Yeah can be removed at this point

hillerstorm:
Set bonus gets applied when running the web sim but not when the tests run, investigating...

PR #651, opened 2024-06-09
[FIX] Removing of Random Suffix and Reforge

1337LutZ:
- Now keeps `randomSuffix` when removing a reforge of a `randomSuffix` item
- Now removes reforge and reforge tab when removing `randomSuffix`
- Added correct button state for `Remove Random Suffix` tab


kayla-glick:
Included in https://github.com/wowsims/cata/pull/648

PR #656, opened 2024-06-10
[UI] Updated Batch Sim UI

kayla-glick:
This is the new Batch Sim UI. Equipped items are shown with the brand-gold border, while new items are shown in grey. Equipped items are not editable from this interface, but any new items can be fully edited. Users can search for items or copy an equipped item to create a new item that can be changed.

image

The updated item search provides far more information for users in a much nicer UI. Updates are immediate and up to 21 results are shown at a time, with a message when there are more results than can be shown. Users can also filter by min and max ilvl in this UI.

image

I cleaned up the Results UI as well.

image

image


ui/core/components/individual_sim_ui/bulk_tab.tsx
```diff
@@ -398,145 +543,141 @@ export class BulkTab extends SimTab {
...
-		const importFavsButton = (
-			<button className="btn btn-secondary w-100 bulk-settings-button">
-				<i className="fa fa-download" /> Import Favorites
-			</button>
-		);
-		importFavsButton.addEventListener('click', () => {
-			const filters = this.simUI.player.sim.getFilters();
-			const items = filters.favoriteItems.map(itemID => ItemSpec.create({ id: itemID }));
-			this.addItems(items);
+		new BooleanPicker<BulkTab>(booleanSettingsContainer, this, {
```
1337LutZ, line 559: To give more control over where you want to append it, you can pass the container as `null` and then actually:
```
const fastModeCheckbox = new BooleanPicker<BulkTab>(null, this, {})

settingsContainer.appendChild(<>
  {fastModeCheckbox.rootElem}  
  {defaultGemDiv}
</>)
```
etc. That way the order wouldn't matter of how we create all these pickers etc.

ui/core/components/individual_sim_ui/bulk_tab.tsx
```diff
@@ -398,145 +543,141 @@ export class BulkTab extends SimTab {
...
+			},
+		});
 
-			if (!searchString.length) {
-				searchResultsRef.value?.replaceChildren();
-				searchResultsRef.value?.classList.add('hide');
-				return;
-			}
+		settingsContainer.appendChild(defaultGemDiv);
+		settingsContainer.appendChild(talentsToSimDiv);
```
1337LutZ, line 632: This can be merged:
```
settingsContainer.appendChild(<>
  {defaultGemDiv}
  {talentsToSimDiv}
</>);
```

ui/scss/core/components/individual_sim_ui/_bulk_tab.scss
```diff
@@ -1,108 +1,81 @@
...
 		}
+	}
 
-		.input-root {
-			flex-direction: row;
-			align-items: center;
-			padding-bottom: var(--block-spacer);
-			border-bottom: var(--border-default);
-			margin-bottom: 0;
+	#bulkSetupTab {
```
1337LutZ, line 13: This doesn't need to be nested, since an id has the highest specificity already.

1337LutZ (changes requested):

1337LutZ (approved):

ui/scss/core/components/individual_sim_ui/_bulk_tab.scss
kayla-glick: I mostly just do this for organization 😅 I like to group the selectors similarly to how they're used on the page just to make them easier to find

PR #659, opened 2024-06-11
[UI] Suggest reforge changes

1337LutZ:
- Added suggest reforge context menu to allow custom EP / stat caps
- Added `useCustomEPValues` property to allow the player to override defaults.
- Added a `statCaps` to the player to manage default hard caps.
- Added action button group
- Added loading state to action buttons
- Added loading state to auto reforge
- Moved logging to only trigger in `devMode`
- Added Arms auto reforge option
- Added Fury auto reforge option (Validated through Fight Club Fury channel with new EPs)

image
image
image


jjayrex (commented):
Can't we do a race check to see if the character is a Draenei and in that case set the Hit rate cap to 7 instead, to account for the Heroic Presence?

1337LutZ:
> Can't we do a race check to see if the character is a Draenei and in that case set the Hit rate cap to 7 instead, to account for the Heroic Presence?

All of that should be tackled by the characterStats functions in the Player class. This is always taking all character buffs into account.

NerdEgghead (approved):

PR #670, opened 2024-06-13
[UI][Reforge Optimizer] Add pre-return stats modifier function

1337LutZ:
- Add a function that allows for stats modifications before they are returned (For spec specific Glyphs etc)


kayla-glick (approved):
Approving but I'll let @NerdEgghead take a look since he wrote the reforger

hillerstorm:
> Approving but I'll let @NerdEgghead take a look since he wrote the reforger

image
🙏 

looks like there's conflicts @1337LutZ

PR #679, opened 2024-06-14
Soft cap breakpoints support for auto-Reforge tool

NerdEgghead:
- Added support for configuring multiple breakpoints for stats with discrete soft caps
- Enabled auto-Reforging for Demonology sim as an example of how to configure soft cap breakpoints


NerdEgghead:
@1337LutZ After messing around with this, I don't think any UI configuration implementation is necessary for breakpoints, and the setup is better left to sim experts for each spec. If a user wants to target a specific stat breakpoint that isn't covered by the default configuration, they can simply run the tool once, then run it again with a manually specified cap equal to the closest breakpoint from their first run, rather than needing to set up a whole list of breakpoint values. But if you think there's a strong use case for having this being fully customizable in the UI, then it can be refactored for that.

NerdEgghead:
@Tharre Let me know if the Demo implementation looks okay to you, I did verify that the tool is resulting in a DPS increase both for the example setup you sent me and for the Phase 1 preset. I didn't put in any Haste breakpoints and only did the discrete Mastery % breakpoints (until they're fixed in-game), because as far as I can tell from direct simming, targeting Haste caps isn't actually real, and Haste looks to be just straight worse than Mastery like your default stat weights indicate.

1337LutZ:
> @1337LutZ After messing around with this, I don't think any UI configuration implementation is necessary for breakpoints, and the setup is better left to sim experts for each spec. If a user wants to target a specific stat breakpoint that isn't covered by the default configuration, they can simply run the tool once, then run it again with a manually specified cap equal to the closest breakpoint from their first run, rather than needing to set up a whole list of breakpoint values. But if you think there's a strong use case for having this be fully customizable in the UI, then it can be refactored for that.

Having the breakpoints in code only might actually be fine, since they are basically not going to shift anyways.

I do still think that perhaps showing these breakpoints somewhere could be good, or that they are implemented for the spec?
What would you think of extending the config for breakpoints to have:
```const masterySoftCapConfig = {
  stat: Stat.StatMastery,
  breakpoints: masteryRatingBreakpoints,
  tooltip: (
    <>Tooltip description here that breakpoints are implemented</>
  ),
};
new ReforgeOptimizer(this, { softCapsConfig: [masterySoftCapConfig] });
```

That way we can add a tooltip on the Suggest button and have it list the breakpoints per spec if they are defined.

NerdEgghead:
> > @1337LutZ After messing around with this, I don't think any UI configuration implementation is necessary for breakpoints, and the setup is better left to sim experts for each spec. If a user wants to target a specific stat breakpoint that isn't covered by the default configuration, they can simply run the tool once, then run it again with a manually specified cap equal to the closest breakpoint from their first run, rather than needing to set up a whole list of breakpoint values. But if you think there's a strong use case for having this be fully customizable in the UI, then it can be refactored for that.
> 
> Having the breakpoints in code only might actually be fine, since they are basically not going to shift anyways.
> 
> I do still think that perhaps showing these breakpoints somewhere could be good, or that they are implemented for the spec? What would you think of extending the config for breakpoints to have:
> 
> ```
>   stat: Stat.StatMastery,
>   breakpoints: masteryRatingBreakpoints,
>   tooltip: (
>     <>Tooltip description here that breakpoints are implemented</>
>   ),
> };
> new ReforgeOptimizer(this, { softCapsConfig: [masterySoftCapConfig] });
> ```
> 
> That way we can add a tooltip on the Suggest button and have it list the breakpoints per spec if they are defined.

Makes sense to me, do you want to apply your UI wizardry to implement this? Can either push to this PR or make a new one after merging this, whatever works better for you!

1337LutZ:
@Tharre if you play around with this, perhaps better to use this PR for your Warlock implementation so you don't have to do it later. https://github.com/wowsims/cata/pull/680

1337LutZ (commented):
Created this PR https://github.com/wowsims/cata/pull/680 to allow the settings to be imported to the BE for CLI usage + add a tooltip to notify users which stat breakpoints have been implemented

Tharre:
> @Tharre Let me know if the Demo implementation looks okay to you, I did verify that the tool is producing a DPS increase both for the example setup you sent me and for the Phase 1 preset. I didn't put in any Haste breakpoints and only did the discrete Mastery % breakpoints (until they're fixed in-game), because as far as I can tell from direct simming, targeting Haste caps isn't actually real, and Haste looks to be just straight worse than Mastery like your default stat weights indicate.

I can't reproduce being below haste cap to be a DPS increase; not sure what's up with that. The default preset is scuffed yeah, I put the heroic wand in it originally and haven't adjusted it since that got removed.

The implementation looks good to me though, I can add the haste caps myself.

NerdEgghead:
> > @Tharre Let me know if the Demo implementation looks okay to you, I did verify that the tool is producing a DPS increase both for the example setup you sent me and for the Phase 1 preset. I didn't put in any Haste breakpoints and only did the discrete Mastery % breakpoints (until they're fixed in-game), because as far as I can tell from direct simming, targeting Haste caps isn't actually real, and Haste looks to be just straight worse than Mastery like your default stat weights indicate.
> 
> I can't reproduce being below haste cap to be a DPS increase; not sure what's up with that. The default preset is scuffed yeah, I put the heroic wand in it originally and haven't adjusted it since that got removed.
> 
> The implementation looks good to me though, I can add the haste caps myself.

I'm going to merge this for now so it's not blocking any specs. Will see if I can figure out that Haste discrepancy vs. your results later, but in any case, the functionality should be set up for you to add Haste breakpoints yourself.

PR #691, opened 2024-06-14
add new melee special outcome that only checks block and crit

zoristaken:
fix warrior raging blow hit outcome
fix rogue mutilate hit outcome
fix shaman stormstrike hit outcome
update tests


sim/warrior/fury/raging_blow.go
```diff
@@ -60,7 +60,7 @@ func (war *FuryWarrior) RegisterRagingBlow() {
 		},
 
 		ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
-			result := spell.CalcOutcome(sim, target, spell.OutcomeMeleeWeaponSpecialHitAndCrit)
+			result := spell.CalcOutcome(sim, target, spell.OutcomeMeleeSpecialHit)
```
rosenrusinov, line 63: Hit and Crit?

rosenrusinov: oh its just a hit check no damage

PR #699, opened 2024-06-15
Add a way for dot spells to be connected to their impact spells

rosenrusinov:
Makes it possible to use the multidot action with dot spells that have their dot component as separate


rosenrusinov:
@Tharre I dont know why warlock tests got changed. Maybe you can spot something but from what code I changed I did not expect any of the test results to change. Shaman and Mage both show same results as before this change but yet warlock does not. I wont merge this PR until we figure out what causes it.

raethkcj:
Thanks for doing this, looks good from a Shaman PoV! Confirmed it works as expected in the Multi Dot action which was my original reason for asking. 

One relatively minor thing is that now both the Direct and DoT spells are shown in in "DoT Spell" APL fields, e.g. DoT Remaining Time:
image

Selecting either one works fine, since they now both refer to the same DoT. Not sure which one makes more sense to keep.

Tharre (commented):
It's because the way `spell.Dot()` got changed; it causes the shadow bite code in pets.go to double count spells like shadowflame and immolate that have a direct damage and a dot component.

I'm not sure if having `.Dot()` return the dot even on the direct spell is really what you'd want, but if so you'd need to create another of those functions that is a copy of the original one to use in the shadowbite code.

sim/warlock/pets.go
```diff
@@ -213,7 +213,7 @@ func (pet *WarlockPet) registerShadowBiteSpell() {
 
 			activeDots := 0
 			for _, spell := range pet.Owner.Spellbook {
-				if spell.Dot(target) != nil && spell.Dot(target).IsActive() {
+				if spell.ClassSpellMask&WarlockDoT > 0 && spell.Dot(target).IsActive() {
```
Tharre, line 216: That won't work, shadow bite scales with all dots, even ones from trinkets and such.

rosenrusinov: I was just about to ask you if it should scale as I saw a test with a dot trinket was still failing.

rosenrusinov: I can do something else which is a bit ugly but should work as expected:
```
if spell.Dot(target) != nil && spell.RelatedDotSpell == nil && spell.Dot(target).IsActive() {
	activeDots++
}
```

This would make it so spells that have a related dot spell are not counted as their related dot spell will be counted anyway

Tharre: Guess that works, gotta add a comment there though, no way someone reading this code understands why it's there 😂

rosenrusinov:
@raethkcj fixed the dot menus to show one of the 2 spells that have the dot linked.

PR #704, opened 2024-06-15
[Warrior] Fix oh proc mask and update dual wield spec mod

zoristaken:
- update warrior oh proc mask
- update dual wield spec to use oh proc mask
- update tests

### DIFF
--- a/sim/warrior/whirlwind.go
+++ b/sim/warrior/whirlwind.go
@@ -18,8 +18,8 @@ func (warrior *Warrior) RegisterWhirlwindSpell() {
 		whirlwindOH = warrior.RegisterSpell(core.SpellConfig{
 			ActionID:       actionID.WithTag(1),
 			SpellSchool:    core.SpellSchoolPhysical,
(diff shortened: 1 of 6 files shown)


sim/warrior/whirlwind.go
```diff
@@ -18,8 +18,8 @@ func (warrior *Warrior) RegisterWhirlwindSpell() {
 		whirlwindOH = warrior.RegisterSpell(core.SpellConfig{
 			ActionID:       actionID.WithTag(1),
 			SpellSchool:    core.SpellSchoolPhysical,
-			ProcMask:       core.ProcMaskEmpty, // whirlwind offhand hits usually don't proc auras
-			ClassSpellMask: SpellDualWieldSpecMask,
+			ProcMask:       core.ProcMaskMeleeOHSpecial, //TODO: needs testing to check if it procs auras, according to pre-cata it didn't
```
rosenrusinov, line 21: If this comment turns out correct then you can add a 2nd class spell mask for WW oh spell and apply a 2nd mod with the same bonus just for it. If after testing WW OH hits should proc stuff then you are already covered.

PR #708, opened 2024-06-15
[FIX] Preset warnings trigger

1337LutZ:
- Fixes reverse DK DW/2H check
- Split message into multiple message per type of warning


ui/core/preset_utils.tsx
```diff
@@ -150,7 +150,7 @@ export const makeSpecChangeWarningToast = (checks: SpecCheckWarning[], player: P
 					))}
 				</>
 			),
-			delay: 5000,
+			delay: 5000 * messages.length,
```
1337LutZ, line 153: Added this so people have time to read as text size increases.

PR #722, opened 2024-06-17
[Shaman] Fix flametongue imbue being affected by reforges

Polynomix:
Also added a test to not have it stack with itself when imbued on both weapon

Not sure why enh tests are not failing


rosenrusinov (approved):
Test are not changing probably because this glyph is not part of the test loadout so its not really covered by tests

Issue #730, opened 2024-06-18
Crash Report 1569997388


Chertus:
Link:
http://localhost:3333/cata/hunter/survival/#eJzlUk9IFFEcnu/31mn2YfgcCsa97OxkMAjF29kOWwgzetqjR7sZmHiIkE51U7qoBEVGrnsIohIJiq4GQQTZP9tZ9uBKQoFpeBBK8OAhpDc99WCnJTr1nb7v+33v/X7v8eMnLQhEGMYIxoEyMAe8BJaBDWCUME54RuhNWyjpiG1kxswKM/suXbh28YqVFpT5Bm6Lmw3mbLme+XgmtTXN/EfgrWK3zDy2rVQFvE0sVZmz6WpDlblYm2fOByvhbWJb8S9ZXSyDHxVVZXx2PRj+bCIbSt7tSOStJL4yzzxTZffuskX1Rsq55x54s0mosaAaZg8atoqf+/P8oVLi6xrjR8TrBeZPQJHKU+Zcz3EjGTNmTs31p+Fxi+TxPuqngZauJ8R3YPynsGWzJ84ZvTSCSUh+mp8Q2cwxiwSKLEKJDbOrevcop7M/Qon2j4sJNsPn1C0L8jBOBQUZHPZkIVC2SifFIB/kC/n8K+q0bsckHqyT83Cd3KmY/HJM8k5MxUadopU6ld7X6S1ZNeKfyFgl7sAz7RQnbmzQ4BjTMxWj+3vsfDTJDG+R0lpejto1ORN1VGYSvAvPaqcW5oZ+Yyns2X9OMEVuv9PS9Y93x5+ImomPDr5pKv830J/z4nvUo/VOOIBf8lqiiw==

RNG Seed: 1964005605

no base result for equipped gear found in bulk sim

1337LutZ:
@Chertus Closing this. This has been fixed in the latest Web sim release. Todays local sim version (0.0.22 deployed later today) should also contain these fixes.

Issue #728, opened 2024-06-17
Crash Report 6223076


gkane1990:
Link:
[...]

Issue #711, opened 2024-06-15
Crash Report -2145705395


AxPython:
Link:
https://wowsims.github.io/cata/paladin/retribution/#eJzlkj9oFEEUxvd7b3O5jELGI8V4ILcuCktA2NtrYgp3kyplytgpxGBhIXZ2l+6ioodXXHKFcISAIMFgFU78s0HQwEWPK6KIIoiaQ0UFiUaI4OxuUmh3iJUfU7z5zffNGx4jDqchEeA0zqIEzALXgfvAM6ADFAklwhJhVEgMYQwT2pgxso97apwaP3Py/KlzOm5m30BkZOMpqy+WnboxZ5Zn2alBmPJWi0W//LbN6r1l86cqO2UIIb+2WH3PRZaMvPuT1aucndKHm9U4JuXlB6x+5GxTM8O5GJF72vVBk82IVKEvXdrgOLRzab+8rftfs3aZBu0NVtu5pO00RK9cXGanArFXvqibCb4a4WKHnYWoWF2O+wu58oTV2gGnDmGIPjn/jmO3vScNV43rIZSBwUUSWzD+U2XcbhPDxqjRQCprClJo0DHPLbiu5/6mI3+CGHoF7fXyemlFoUI+v0KH0jffkny5TupKSFYpJKcSkjsT0tClkIJOm8aabXpEZot6n5PxmnTXDk1Oc/KagaC+Ux0PLrBhN6kv3hWngn27jv21uUir/tGEtPyDU7HW/ZG1ZqSPvlcha0L1DP7jn+DMBN3Yi5MPu/L/jZLh3PkcjCT7Lf8EfgHXsJjq

RNG Seed: 1504668154

simulation failed: No random suffix with id: -225
Stack Trace:
goroutine 21 [running]:
runtime/debug.Stack()
	/opt/hostedtoolcache/go/1.21.11/x64/src/runtime/debug/stack.go:24 +0x6
github.com/wowsims/cata/sim/core.runSim.func1()
	/home/runner/work/cata/cata/sim/core/sim.go:124 +0x10
panic({0x107ac0, 0x395aa90})
	/opt/hostedtoolcache/go/1.21.11/x64/src/runtime/panic.go:914 +0x2c
github.com/wowsims/cata/sim/core.NewItem({0xff5a, 0xffffff1f, 0x0, {0x3312100, 0x2, 0x2}, 0x0})
	/home/runner/work/cata/cata/sim/core/database.go:343 +0x35
[...]

PR #741, opened 2024-06-18
[FIX] Bulk sim issues

1337LutZ:
- Added missing `randomSuffixes` and `reforgeStats` to database
  - https://github.com/wowsims/cata/issues/711
  - https://github.com/wowsims/cata/issues/728
  - https://github.com/wowsims/cata/issues/728
- Fix sim breaking when duplicate equipped items were detected (non-unique)
  - https://github.com/wowsims/cata/issues/730
- Reset pending state on error

### DIFF
--- a/sim/core/bulksim.go
+++ b/sim/core/bulksim.go
@@ -495,23 +495,8 @@ func isValidEquipment(equipment *proto.EquipmentSpec) bool {
 		return false
 	}
 
-	// Validate trinkets/rings for duplicate IDs
-	if equipment.Items[proto.ItemSlot_ItemSlotFinger1].Id == equipment.Items[proto.ItemSlot_ItemSlotFinger2].Id && equipment.Items[proto.ItemSlot_ItemSlotFinger1].Id != 0 {
-		return false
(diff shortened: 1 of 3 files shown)


sim/core/bulksim.go
```diff
@@ -495,23 +495,8 @@ func isValidEquipment(equipment *proto.EquipmentSpec) bool {
 		return false
 	}
 
-	// Validate trinkets/rings for duplicate IDs
-	if equipment.Items[proto.ItemSlot_ItemSlotFinger1].Id == equipment.Items[proto.ItemSlot_ItemSlotFinger2].Id && equipment.Items[proto.ItemSlot_ItemSlotFinger1].Id != 0 {
```
1337LutZ, line 499: There's no need to do this, as the FE UI doesn't allow for unique-equipped duplicates, and this would return as not valid if there's items that are not unique-equipped.

sim/core/bulksim.go
```diff
@@ -495,23 +495,8 @@ func isValidEquipment(equipment *proto.EquipmentSpec) bool {
...
-		return false
-	} else if equipment.Items[proto.ItemSlot_ItemSlotTrinket1].Id == equipment.Items[proto.ItemSlot_ItemSlotTrinket2].Id && equipment.Items[proto.ItemSlot_ItemSlotTrinket1].Id != 0 {
-		return false
-	}
-
-	// Validate rings/trinkets for heroic/non-heroic (matching name)
```
1337LutZ, line 505: Same here, no need to check all of this.

PR #784, opened 2024-06-25
Remove random suffix and reforge from 60U

1337LutZ:
Added https://github.com/wowsims/cata/issues/785 to improve random suffix handling by allowing to migrate them to the proper versions if they are applicable to an item.

When importing from 60U the wrong random suffix will be given. I remove the suffix + reforge to prevent errors. When this happens we inform the user:
image

### DIFF
--- a/ui/core/components/importers.tsx
+++ b/ui/core/components/importers.tsx
@@ -285,10 +285,12 @@ export class Individual60UImporter<SpecType extends Spec> extends Importer {
 			if (itemJson.gems) {
 				itemSpec.gems = (itemJson.gems as Array<any>).filter(gemJson => gemJson?.id).map(gemJson => gemJson.id);
 			}
+			if (itemJson.suffixId) {
+				itemSpec.randomSuffix = itemJson.suffixId;
+			}
 			if (itemJson.reforge?.id) {
 				itemSpec.reforging = itemJson.reforge.id;
 			}
-
 			equipmentSpec.items.push(itemSpec);
 		});
 
--- a/ui/core/components/importers.tsx
+++ b/ui/core/components/importers.tsx
@@ -275,6 +275,8 @@ export class Individual60UImporter<SpecType extends Spec> extends Importer {
 			talentsStr = talentSpellIdsToTalentString(charClass, talentIds);
 		}
 
+		let hasRemovedRandomSuffix = false;
+		const modifiedItemNames: string[] = [];
 		const equipmentSpec = EquipmentSpec.create();
 		(importJson.items as Array<any>).forEach(itemJson => {
 			const itemSpec = ItemSpec.create();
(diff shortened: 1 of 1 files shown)


ui/core/components/importers.tsx
```diff
@@ -285,10 +285,12 @@ export class Individual60UImporter<SpecType extends Spec> extends Importer {
 			if (itemJson.gems) {
 				itemSpec.gems = (itemJson.gems as Array<any>).filter(gemJson => gemJson?.id).map(gemJson => gemJson.id);
 			}
+			if (itemJson.suffixId) {
+				itemSpec.randomSuffix = itemJson.suffixId;
```
InDebt, line 289: Can we validate the suffix ID in browser?
Right now this will lead to a huge number of crash reports as I have yet to see a correct random suffix ID from 60 upgrades. Haven't testing manually setting it but no random suffix ID is correct when linking your character to your battle.net character.

I have already posted that in the cata bug channel on their discord.

1337LutZ: I just tested it with the provided import string using `-116` and that did work some of the other ones such as `-230` (feverflare) also work, and they match with the list we validate against in backend.

Sure we can add `suffixId` and `reforge` testing, but then we would also need to add a toast to inform the user to double check their gear.

InDebt: Intersting.

For example this belt https://www.wowhead.com/cata/item=63498/soul-breath-belt 60 upgrades will always give me -186 as the random suffix ID which is Feverflare and +154 Master and Haste. It should be 160 Mastery and Haste however.

So I assume it's just certain Items that are buged and not random suffixes in general. As even manually selecting the suffix will result in -186

InDebt: Our own UI is using ID -213 for feverflare for this belt. They can't both be right, can they? 🤔

1337LutZ: Just checked our DB:
```
{"id":-213,"name":"of the Feverflare","stats":[0,0,7889,4834,0,0,0,0,0,3320,0,0,0,0,3320,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,3320]},
{"id":-186,"name":"of the Feverflare","stats":[0,0,7889,4638,0,0,0,0,0,3205,0,0,0,0,3205,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,3205]},
```
They are both "valid" IDs, but the stats differ a bit. However it looks like the wowhead tooltip code actually has a way of "migrating" to the correct IDs.
Problem being that our code will not do that and indeed select the wrong stats. I think I should for now actually remove any  `reforge` if `randomSuffix` is set, and show a Toast after importing that we did so.

rosenrusinov: Isnt this just “normal” and “heroic” versions of the random suffix? They should both be valid.

InDebt: Well an easy migration strategy would be to map them based on the name. There will never be two ID's with the same name valid for the same Item. So the UI seems to know which IDs are valid for an item as the selection only shows 4 valid ones.

Thus we could do `ID -> Lookup Name -> Lookup ID for Item based on Name`

1337LutZ: @InDebt Added the following Toast + removed the random suffix + reforge from 60U: 
image

1337LutZ: > Well an easy migration strategy would be to map them based on the name. There will never be two ID's with the same name valid for the same Item. So the UI seems to know which IDs are valid for an item as the selection only shows 4 valid ones.
> 
> Thus we could do `ID -> Lookup Name -> Lookup ID for Item based on Name`

We probably also need to migrate the reforges then right? I think for now showing this message and then in time create a proper migration would be the best way forward. @rosenrusinov @InDebt

1337LutZ: We should also update our `itemSpec` logic in `database.ts` because that just gets the suffix by ID no questions asked:
```
let randomSuffix: ItemRandomSuffix | null = null;
if (itemSpec.randomSuffix && !!this.getRandomSuffixById(itemSpec.randomSuffix)) {
	randomSuffix = this.getRandomSuffixById(itemSpec.randomSuffix)!;
}
```

1337LutZ:
Created https://github.com/wowsims/cata/issues/785 to add the random suffix improvements

PR #786, opened 2024-06-25
Makefile edit for FreeBSD support

dexxiez:
Make file edit to check if there was a change on package.json and if the user is running FreeBSD. If they are it will use rollup instead of Vite due to incompatibilities. It will run `npm install` after to fix any lock file inconsistencies. 

I chose `jq` instead of any build in util like `sed` etc because `jq` is structure aware instead of straight text find and replace.

addresses #586

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,8 @@ sim/web/__debug_bin
 
 # temporary files
 *.results.tmp
+temp.json
+
 #.dockerignore
 /cata/
 *.code-workspace
--- a/makefile
+++ b/makefile
@@ -95,8 +95,8 @@ ui/%/index.html: ui/index_template.html
 .PHONY: package.json
 
 package.json:
-# Checks if the system is FreeBSD and jq is installed. This is due to the need to switch out the rollup package for vite.
-ifeq ($(shell uname -s), FreeBSD)
+# Checks if the system is FreeBSD and jq is installed. This is due to the need to switch out the vite package for rollup on FreeBSD.
+ifeq ($(shell uname -s), Linux)
 	@if ! command -v jq > /dev/null; then \
 		echo "jq is not installed. Please install jq to proceed."; \
 		exit 1; \
--- a/makefile
+++ b/makefile
@@ -92,6 +92,25 @@ ui/core/proto/api.ts: proto/*.proto node_modules
 ui/%/index.html: ui/index_template.html
 	cat ui/index_template.html | sed -e 's/@@CLASS@@/$(shell dirname $(@D) | xargs basename)/g' -e 's/@@SPEC@@/$(shell basename $(@D))/g' > $@
 
+.PHONY: package.json
+
+package.json:
+# Checks if the system is FreeBSD and jq is installed. This is due to the need to switch out the rollup package for vite.
+ifeq ($(shell uname -s), FreeBSD)
+	@if ! command -v jq > /dev/null; then \
+		echo "jq is not installed. Please install jq to proceed."; \
+		exit 1; \
+	fi; \
+	\
+	echo "Checking and updating package.json for FreeBSD..."; \
+	\
+	if ! grep -q '"overrides"' package.json; then \
+		jq '. + { "overrides": { "vite": { "rollup": "npm:@rollup/wasm-node@4.13.0" } } }' package.json > temp.json && mv temp.json package.json && npm install; \
+	else \
+		jq '.overrides += { "vite": { "rollup": "npm:@rollup/wasm-node@4.13.0" } }' package.json > temp.json && mv temp.json package.json && npm install; \
+	fi
+endif
+
 package-lock.json:
 	npm install
 


1337LutZ:
How will this revert the change when you want to commit? Or is it up to the user to not commit the changes?

dexxiez:
> How will this revert the change when you want to commit? Or is it up to the user to not commit the changes?

I was considering how to approach this, as within the issue where I put in git statements, I didn't know if it would be appropriate to put git related items in a makefile, bit I might be thinking too much into it, I don't know. Thoughts?

1337LutZ:
The easiest solution would be to just keep track of the `package.json` when people create pull requests.
I have had the idea for a while to add [Husky](https://typicode.github.io/husky/)  for git hooks which could tackle that. However would prefer some more time to set it up.

1337LutZ:
@dexxiez 
Actually another one could be creating a test in the test script that just checks if the package.json contains the wasm override and let it fail

dexxiez:
> @dexxiez Actually another one could be creating a test in the test script that just checks if the package.json contains the wasm override and let it fail

Smart. ill have a squiz. one sec

.github/workflows/run_tests.yml
```diff
@@ -1,38 +1,45 @@
...
+              run: go install google.golang.org/protobuf/cmd/protoc-gen-go@latest
 
-      - name: Install Node
-        uses: actions/setup-node@v3
-        with:
-          node-version: '19'
+            - name: Install Node
+              uses: actions/setup-node@v3
+              with:
+                  node-version: 'lts/*'
```
1337LutZ, line 27: This should be changed to 20, not LTS since everything locally is now documented as 20

.github/workflows/release.yml
```diff
@@ -1,69 +1,68 @@
...
+              run: go install google.golang.org/protobuf/cmd/protoc-gen-go@latest
 
-      - name: Install Node
-        uses: actions/setup-node@v3
-        with:
-          node-version: '19'
+            - name: Install Node
+              uses: actions/setup-node@v3
+              with:
+                  node-version: 'lts/*'
```
1337LutZ, line 39: This should be changed to 20, not LTS since everything locally is now documented as 20

.github/workflows/deploy.yml
```diff
@@ -1,45 +1,45 @@
...
+              run: go install google.golang.org/protobuf/cmd/protoc-gen-go@latest
 
-      - name: Install Node
-        uses: actions/setup-node@v3
-        with:
-          node-version: '19'
+            - name: Install Node
+              uses: actions/setup-node@v3
+              with:
+                  node-version: 'lts/*'
```
1337LutZ, line 27: This should be changed to 20, not LTS since everything locally is now documented as 20

.github/workflows/update_items.yml
```diff
@@ -1,64 +1,63 @@
...
+              run: go install google.golang.org/protobuf/cmd/protoc-gen-go@latest
 
-      - name: Install Node
-        uses: actions/setup-node@v3
-        with:
-          node-version: '19'
+            - name: Install Node
+              uses: actions/setup-node@v3
+              with:
+                  node-version: 'lts/*'
```
1337LutZ, line 29: This should be changed to 20, not LTS since everything locally is now documented as 20

1337LutZ (changes requested):

.github/workflows/run_tests.yml
dexxiez: We can explicitly set it to 20 however my reasoning behind picking the latest LTS is:

A - run into the same thing that this replaces which is changing the engine version in package json but still keeping the previous (or previous previous) version in CICD because it hasn't needed to be looked at for a while.

B - If there is something that doesn't run on the latest stable LTS, that probably is a problem which should be addressed. 

This syntax does resolve the version currently to 20.4.0 


Thoughts?

1337LutZ: In rare occasions a mismatch in Node versions could introduce code that doesn't work anymore. It's rare, but happens.
So I prefer to set it to `20.15.0` which is the current 20.x LTS, and then make sure we check more precise when upgrading node versions.

dexxiez: Changed to 20 now for `20.x`. Im working out a way to future proof the engines and node mismatch for my cicd pipelines so I'll probably end up making a quick github action one time to check it is compliant with the engines field or it will fail. It will be useful for projects that need to use npm (like electron projects). I tried to look at using `engine-strict` however it only works for upstream packages and not the root package.json you're using it on. It's a bit of a pain as pnpm has this functionality, and I think it makes sense to not implement it in npm but 🤷

1337LutZ (approved):

.gitignore
```diff
@@ -18,6 +18,8 @@ sim/web/__debug_bin
 
 # temporary files
 *.results.tmp
+temp.json
```
kayla-glick, line 21: What do you think about naming something like `package-tmp.json?` That was it's more clear that this is a copy of a `package.json` file?

MikeJakubik:
This is great, the cherry on the top would be to not rely on jq however, but something more standard (i.e. sed, awk, etc.). Jq is not part of a FreeBSD install, but it is available as a port/pkg. 

However, I understand you want to work with more parsable data types, FreeBSD does have a neat feature as part of its base install, and that is [libxo](https://github.com/Juniper/libxo). Many commands support it directly or you can use the xo command to work with data structures such as XML, Json, or HTML. I'm curious if this would be usable here, but if you have a working solution and no time then it's OK, just thought it would be neat to share this info.

[libxo (man page)](https://man.freebsd.org/cgi/man.cgi?query=libxo)


[xo (command man page)](https://man.freebsd.org/cgi/man.cgi?query=xo)


Here are some neat examples

```
[root@fbsd14 ~]# wc --libxo json,pretty,warn /etc/motd
{
  "wc": {
    "file": [
      {
        "lines": 21,
        "words": 97,
        "characters": 900,
        "filename": "/etc/motd"
      }
    ]
  }
}

[root@fbsd14 ~]# ps auxw -U _dhcp --libxo json,pretty,warn
{
  "process-information": {
    "process": [
      {
        "user": "_dhcp",
        "pid": "1886",
        "percent-cpu": "0.0",
        "percent-memory": "0.0",
        "virtual-size": "13304",
        "rss": "2960",
        "terminal-name": "- ",
        "state": "ICs",
        "start-time": "22:31  ",
        "cpu-time": "0:00.00",
        "command": "dhclient: wlan0 (dhclient)"
      }
    ]
  }
}
```
In a scripted environment

```
#!/bin/sh
xo --pretty --style json --top-wrap --open hello/user '{:name} {:date}' "$(whoami)" "$(date +%H:%M:%S)"
xo --pretty --style json --top-wrap --close hello/user
```

```
[root@fbsd14 ~]# sh hello.sh 
{
  "hello": {
    "user": {
      "name": "root",
      "date": "22:30:06"
    }
  }
}
```

dexxiez:
> This is great, the cherry on the top would be to not rely on jq however, but something more standard (i.e. sed, awk, etc.). Jq is not part of a FreeBSD install, but it is available as a port/pkg.
> 
> However, I understand you want to work with more parsable data types, FreeBSD does have a neat feature as part of its base install, and that is [libxo](https://github.com/Juniper/libxo). Many commands support it directly or you can use the xo command to work with data structures such as XML, Json, or HTML. I'm curious if this would be usable here, but if you have a working solution and no time then it's OK, just thought it would be neat to share this info.
> 
> [libxo (man page)](https://man.freebsd.org/cgi/man.cgi?query=libxo)
> 
> [xo (command man page)](https://man.freebsd.org/cgi/man.cgi?query=xo)
> 
> Here are some neat examples
> 
> ```
> [root@fbsd14 ~]# wc --libxo json,pretty,warn /etc/motd
> {
>   "wc": {
>     "file": [
>       {
>         "lines": 21,
>         "words": 97,
>         "characters": 900,
>         "filename": "/etc/motd"
>       }
>     ]
>   }
> }
> 
> [root@fbsd14 ~]# ps auxw -U _dhcp --libxo json,pretty,warn
> {
>   "process-information": {
>     "process": [
>       {
>         "user": "_dhcp",
>         "pid": "1886",
>         "percent-cpu": "0.0",
>         "percent-memory": "0.0",
>         "virtual-size": "13304",
>         "rss": "2960",
>         "terminal-name": "- ",
>         "state": "ICs",
>         "start-time": "22:31  ",
>         "cpu-time": "0:00.00",
>         "command": "dhclient: wlan0 (dhclient)"
>       }
>     ]
>   }
> }
> ```
> 
> In a scripted environment
> 
> ```
> #!/bin/sh
> xo --pretty --style json --top-wrap --open hello/user '{:name} {:date}' "$(whoami)" "$(date +%H:%M:%S)"
> xo --pretty --style json --top-wrap --close hello/user
> ```
> 
> ```
> [root@fbsd14 ~]# sh hello.sh 
> {
>   "hello": {
>     "user": {
>       "name": "root",
>       "date": "22:30:06"
>     }
>   }
> }
> ```

I've had a look, and it looks like there isn't much parsing functionality in xo. It seems to be more 1 way which is out (emitting), so you use their syntax to emit  other formats, JSON being one of them. So sadly I don't think it can be used easily in this context.

.gitignore
dexxiez: > What do you think about naming something like `package-tmp.json?` That was it's more clear that this is a copy of a `package.json` file?

new day, new me. 

`package.json.tmp` okay?

1337LutZ: @dexxiez Yea that sounds fine. Also that file shouldn't exist after modifying anyways, right?
As in, it uses the `mv` command to make the temp, but then swap it back and run `npm install`

dexxiez: Yeah under normal circumstances it will turn into a normal `package.json`. I'm no low level expert but I think the main reasoning behind the temp file write is to do with stream safety to not directly write back to the original file instantly. 

This line in the ignore file just accounts for glitches in space, time and quantum mechanics etc

1337LutZ: Yea was just double checking I didn't miss anything.

kayla-glick (approved):

PR #789, opened 2024-06-25
[WARR][FURY] Add softcaps for TG and SMF

1337LutZ:
- Recalculated Fury SMF & TG EPs
- Added Fury hit softcap EPs


1337LutZ:
Validated by Falfa in Fight Club Discord

PR #808, opened 2024-07-01
[SHADOW] Add shadow priest softcap breakpoints

1337LutZ:
Calculated and added breakpoints together with InDebt

Based off this detailed scaling graph and picked some minor peaks to allow for in between breakpoints
image

### DIFF
--- a/ui/priest/shadow/presets.ts
+++ b/ui/priest/shadow/presets.ts
@@ -42,9 +42,9 @@ export const P1_EP_PRESET = PresetUtils.makePresetEpWeights(
 		[Stat.StatSpirit]: 0.53,
 		[Stat.StatSpellPower]: 0.82,
 		[Stat.StatSpellHit]: 0.52,
-		[Stat.StatSpellCrit]: 0.49,
+		[Stat.StatSpellCrit]: 0.42,
 		[Stat.StatSpellHaste]: 0.21,
-		[Stat.StatMastery]: 0.46,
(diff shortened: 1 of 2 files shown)


1337LutZ:
Put this on pending to fine tune with Lenlo 👍

1337LutZ:
@InDebt Ready to be reviewed. Refined additional semi-breakpoints to smoothen the EP scaling a bit. During that found some additional points.

InDebt:
Those breakpoints seem reasonable.
I am wondering about two points though:
* Why those points were plotted without shadow form (according to shadow discord you used 3% talented haste and not buffs?)
* Haste is always stronger than any other stat we can reforge, so is there any benefit in setting those values?

1337LutZ:
I did it Naked because I talked to Linelo and that's how he did all his research. So the Haste breakpoints were done like this:
- Naked (no gear)
- 10k SP
- 999k Mana
- Default Rotation (APL)
- Default Talents
- No buffs

Then for all the EP values I actually used:
- P1 BIS
- Default Rotation (APL)
- Default Talents
- No buffs
- Reduced Hit to 1701
- Start Haste 40 below first breakpoint

Then increment towards each of the values

1337LutZ:
Closing this as we will use default EPs as a default since prioritizing haste is the best no-brain option.
Added dropdown for breakpoint selection is this PR to allow for easier manual fine tuning: https://github.com/wowsims/cata/pull/825

PR #822, opened 2024-07-07
Add filler breakpoints to allow crit prio

1337LutZ:
- Added filler breakpoints to improve Haste x Crit variation
  - Doesn't deadlock 
- Changed to Threshold capping due to how Crit always > Haste when no breakpoint is met. (until you are in the 10000+ crit rating levels)

Red = Breakpoint
Black = Filler

image


1337LutZ:
Did some more testing, and this solution also doesn't work as intended.

PR #825, opened 2024-07-09
[UI][REFORGE] Stat cap default

1337LutZ:
Shadow Priests and Mages have varying Haste + Mastery/Crit interactions which makes it hard to calculate breakpoints effectively for.
Currently they use the hard cap functionality (also in ReforgeLite) to aim for a breakpoint (Haste) value and let the rest be filled based off EP.
This makes having to remember all breakpoints and manually entering them each time in the reforge tool quite time intensive.

Adding a list of relevant breakpoints therefore will allow for quicker reforging based on manual stat caps.

**Example**
image
image


ui/mage/fire/presets.ts
```diff
@@ -41,6 +41,44 @@ export const P1_EP_PRESET = PresetUtils.makePresetEpWeights(
...
+			['16-tick Combust', 7033],
+			['7-tick LvB/Pyro', 8000],
+			['17-tick Combust', 8309],
+			['18-tick Combust', 9602],
+			['19-tick Combust', 10887],
+			['8-tick LvB/Pyro', 11198],
+			['20-tick Combust', 12182],
+			['21-tick Combust', 13463],
+			['9-tick LvB/Pyro', 14412],
+			['22-tick Combust', 14704],
```
InDebt, line 67: Looks good! Mostly aligns with the breakpoints we setup for the softcaps so should yield nice results for manual optimization.

InDebt (approved):

ui/mage/fire/presets.ts
1337LutZ: Yea, I will expand the breakpoints config for soft capping once P2 gets announced

PR #830, opened 2024-07-09
Non-zero post-cap EP support for "Threshold" stats in Reforge optimizer

NerdEgghead:
Added support for specifying non-zero post-threshold EPs for "Threshold" stat types, and changed Fire Mage soft cap configuration to TypeThreshold as an example of how this can be utilized for improved auto-Reforge behavior.

 On branch auto_reforge
 Changes to be committed:
	modified:   package-lock.json
	modified:   ui/core/components/suggest_reforges_action.tsx
	modified:   ui/mage/fire/sim.ts
	modified:   ui/warlock/affliction/sim.ts
	modified:   ui/warlock/demonology/sim.ts


NerdEgghead:
@1337LutZ Let me know if this change fixes the problematic cases for Fire Mage where users were hacking negative Mastery EPs with manually forced Haste caps to produce the desired output. I did verify that this version produces a higher DPS solution than the prior configuration when run on both the pre-raid and Phase 1 gear presets.

1337LutZ:
> @1337LutZ Let me know if this change fixes the problematic cases for Fire Mage where users were hacking negative Mastery EPs with manually forced Haste caps to produce the desired output. I did verify that this version produces a higher DPS solution than the prior configuration when run on both the pre-raid and Phase 1 gear presets.

Awesome! Yea, this works and produces results much closer to what's needed for Mages. :) I added 1 more breakpoint for during Bloodlust, rest seems good.
Also asked @InDebt to double check it 👍 
Thanks again!

1337LutZ (approved):

PR #845, opened 2024-07-14
[Prot] Add remaining Protection Paladin talents and defensive abilities

hillerstorm:
* Fix base dodge
* Add Divine Protection
* Add Holy Shield
* Add Ardent Defender


psiven (approved):
Looks good to me! Haven't tested it but if that Dynamic func works for AD should be good to go.

PR #866, opened 2024-07-23
Versioned Protos with Data Migration

NerdEgghead:
This PR aims to solve a longstanding issue with our codebase, where changes to the size, indexing, or interpretation of protos completely break the sim for the majority of users, even when these changes are in principle a good idea. To solve this, a version number is introduced for our protos, in the form of a custom option for a placeholder `ProtoVersion` message. The front-end code then compares this version to the stored version number for any protos it tries to load from local storage or links, and automatically runs up-conversion functions on the old data if they do not match.

The functionality is tested against a "trivial but breaking" proto change where the Stat enum is simply re-ordered so that stats are grouped more intelligently. This has zero impact on core sim results (once the database is regenerated using the new schema), but would previously make the browser sim unusable without fully restoring defaults, since saved target armor values etc. would be incorrectly indexed. On this PR branch, local storage data and old sim links are correctly migrated to the new Stat schema upon a reload, making it possible to do more significant stats revisions in the future without adversely affecting our users.


ui/core/proto_utils/stats.ts
```diff
@@ -271,11 +274,47 @@ export class Stats {
...
+	static migrateStatsArray(oldStats: Array<number>, oldApiVersion: number): Array<number> {
+		const conversionMap: ProtoConversionMap<Array<number>> = new Map([
+			[1, (oldArray: Array<number>) => {
+				// Revision 1 simply re-orders the stats for clarity
+				const newIndices = [0, 1, 2, 3, 4, 17, 18, 6, 8, 10, 19, 15, 5, 7, 9, 11, 29, 26, 16, 30, 12, 13, 20, 28, 21, 22, 23, 24, 25, 27, 14];
+				const newArray: Array<number> = new Array(oldArray.length);
+				oldArray.forEach((value, idx) => {
+					newArray[newIndices[idx]] = value;
+				});
+				return newArray;
```
1337LutZ, line 313: Potato-potato, personal preference:
```
return oldArray.reduce<number[]>((newArray, value,idx) => {
	newArray[newIndices[idx]] = value;
	return newArray
}, new Array(oldArray.length))
```

ui/core/encounter.ts
```diff
@@ -182,4 +190,19 @@ export class Encounter {
 			targetInputs: new Array<TargetInput>(0),
 		});
 	}
+
+	static updateProtoVersion(oldProto: EncounterProto): EncounterProto {
+		let migratedProto = oldProto;
```
1337LutZ, line 195: This can be a const since you never reassign the whole object

1337LutZ (approved):
Very nice! This is how I had it in mind as well!

Only question I have is;
Should we also try/catch these migrations? And if failed > reset to defaults and show a Toast that we did so?

ToxicKevinFerm:
Looks great, and easy to extend with new stuff!
Will make larger refactorings safer

PR #918, opened 2024-08-06
Allow Apparatus of Khaz'goroth to be used at any stack size

hillerstorm:


sim/common/cata/other_effects.go
```diff
@@ -761,3 +624,129 @@ func init() {
...
+		labelSuffix := core.Ternary(config.Heroic, " (Heroic)", "")
+		buffDuration := time.Second * 15
+		buffAuraOnGain := func(aura *core.Aura, sim *core.Simulation) {
+			aura.SetStacks(sim, stacks)
+			character.AddStatsDynamic(sim, buffs)
+		}
+		buffAuraOnExpire := func(aura *core.Aura, sim *core.Simulation) {
+			character.AddStatsDynamic(sim, buffs.Invert())
+		}
+
```
NerdEgghead, line 650: I think it will be cleaner to create a shared utility similar to `NewTemporaryStatsAura()`, which we can name something like `NewTemporaryStatBuffWithStacks()`. This helper could take in the label, spellID, max stacks, stat type, and stat bonus per stack, and then automatically register an `OnStacksChange` callback to perform the multiplication on the bonusPerStack and call `AddStatDynamic()` with the stat type. Then you simply register the three types of buff auras below with the helper, and you only need to call `SetStacks()` after `Activate()` within the trinket activation spell, without needing to do all the multiplications here.

Your version works fine, but making it reusable is cleaner in case there are other such trinkets in the future. Doing it this way also means you don't need to rely on outer scope variables for `stacks` and `buffs`, since everything you need will be contained within the Aura structs.

NerdEgghead: In fact, there's already a `MakeStackingAura()` helper in sim/core/aura_helpers.go that you can take advantage of when creating the `NewTemporaryStatBuffWithStacks()` helper.

NerdEgghead (changes requested):

sim/common/cata/other_effects.go
hillerstorm: Something like 4b011af ?

NerdEgghead: > Something like [4b011af](https://github.com/wowsims/cata/commit/4b011afa8846806e964f3dc9bbf970968846231b) ?

Sorry for the delayed review, just came back from vacation. Yes this looks much cleaner now, thanks! Feel free to merge it into `master` now, or alternatively I can merge it into the `stats_refactor` branch and fix the conflicts to ship the two changes together once the refactor is ready. Whichever way you prefer is fine, I'll have to fix merge conflicts on my end either way so doesn't matter to me.

NerdEgghead (approved):

sim/common/cata/other_effects.go
hillerstorm: Seeing as the refactor is ready for review and the trinket still is "unusable" for rets until that one is merged, you can go ahead and merge it into `stats_refactor` and ship both 👍

NerdEgghead: I've merged in the stats refactor today, so I'll now fix this PR and ship it either later today or tomorrow.

NerdEgghead: I created a PR on your fork for merging in `stats_refactor`, once you approve that then I believe this current PR should be merge-able.

PR #931, opened 2024-08-13
[UI] Results table refactor

1337LutZ:
- Tweaked the Results UI to give a clearer overview and hide data in tooltips
- Added `SpellSchool` for spell coloring (and multi spell)
  - I don't have the knowledge to refactor Backend to use the shared Proto and migrate the current Encounter specific `SpellSchool ` proto. (Imho can be done in a later PR)
- Add `CritDamage`, `CritHealing`, `Ticks`, `CritTicks`, `TickDamage`, `CritTickDamage`, `GlanceDamage`, `BlockDamage`, `CritBlockDamage`
- Add missing pet names for `ActionId` matching
- Moved all Threat columns under their specific "regular" metrics into a tooltip
- Removed obsolete DPASP metric
- Fixed `CastTime` metric for Channeled spells

**Todo**
- Create a "compound pet metric" to allow Pet (summon) casts to be separated from the actions to accurately show Avg Hit/Casts and DPET for 1 pet summon.

**Screenshots**

**Tank**
image
image

**DPS**
image
image
image
image


1337LutZ:
Could someone let me know why `BaubleofTrueBlood` in `wotlk/other_effects.go` is the only item that breaks the tests?
I couldn't find anything weird in my code:
```test_result_compare.go:46: RaidSimResult.RaidMetrics.Parties[0].Players[0].Actions[item_id:50726][1].CritHealing: Expected 219687.175842 but is 0.000000 for multi threaded result!```

**Comment by 1337LutZ** [comment]
@rosenrusinov In the code I saw that Bloodworms and Rune of the Fallen Crusader can't crit their heals, however they can;
- https://classic.warcraftlogs.com/reports/wAtVXFd6YbhmLWya#boss=-2&difficulty=0&wipes=2&type=healing&source=37
- https://classic.warcraftlogs.com/reports/wAtVXFd6YbhmLWya#boss=-2&difficulty=0&wipes=2&type=healing&source=16

**Comment by rosenrusinov** [comment]
> @rosenrusinov In the code I saw that Bloodworms and Rune of the Fallen Crusader can't crit their heals, however they can;
> 
> * https://classic.warcraftlogs.com/reports/wAtVXFd6YbhmLWya#boss=-2&difficulty=0&wipes=2&type=healing&source=37
> * https://classic.warcraftlogs.com/reports/wAtVXFd6YbhmLWya#boss=-2&difficulty=0&wipes=2&type=healing&source=16
> 
> Can I just add the default crit modifier and `OutcomeHealingCrit` or is this more involved?

If they work with the standard spell crit when they roll for their critical sure. If not they will require their own implementations of the outcome to calculate the crit. I havent done any testing but from taking a look at the 2 links you shared it doesnt seem to be using the DKs crit chance for them.

File: sim/death_knight/death_strike.go [hunk 1]
```diff
@@ -39,7 +39,7 @@ func (dk *DeathKnight) registerDeathStrikeSpell() {
 		ClassSpellMask: DeathKnightSpellDeathStrikeHeal,
 
 		DamageMultiplier: 1,
-		ThreatMultiplier: 1,
```
rosenrusinov, line 42: Healing usually has a threat modifier. Is this change confirmed?

ui/core/components/detailed_results/aura_metrics.ts
```diff
@@ -52,19 +53,25 @@ export class AuraMetricsTable extends MetricsTable<AuraMetrics> {
...
+		});
 	}
 
 	shouldCollapse(metric: AuraMetrics): boolean {
 		return !metric.unit?.isPet;
 	}
+
+	filterMetrics(metrics: Array<AuraMetrics>): Array<AuraMetrics> {
+		return metrics.filter(aura => aura.actionId.otherId !== OtherAction.OtherActionMove);
+	}
```
rosenrusinov, line 76: Why are you removing the Move metric? I think its important if an encounter has movement in it for the user to be able to see how much time was spent while moving. Or maybe I'm not understanding this and its doing something else

rosenrusinov:
I like how this looks, only thing I don't like is the DPS column. I kinda like the full number being shown there instead of it being abbreviated with a K when above a thousand. It has the same length and just makes the information have 2 visual states instead of just 1 (abbreviated and non abbreviated). I think we should revert to just showing the full dps number.

sim/death_knight/death_strike.go
1337LutZ: @rosenrusinov This was posted by Char here in our Discord, you can check with him; https://discord.com/channels/891730968493305867/909597996449157160/1271941989214191737
Or should I in that case put it to 0 if there's no threat for it?

ui/core/components/detailed_results/aura_metrics.ts
1337LutZ: You are correct, I need to only remove it when it concerns pet. At least that was the ask.

ui/core/proto_utils/sim_result.ts
```diff
@@ -875,6 +965,78 @@ export class ActionMetrics {
...
+			case 31803: // Censure - DoT
+				if (tag === 2) {
+					isProc = true;
+				}
+				break;
+			default:
+				break;
+		}
+		return isProc;
+	}
```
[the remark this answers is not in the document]
1337LutZ: @rosenrusinov I asked the question here; https://discord.com/channels/891730968493305867/909597996449157160/1272471428058320926

But good to know. Do you have a preference for naming / the implementation to make it clear that something is a passive "proc" of another ability that counts as a cast?
ie. a `IsProc` spell should never have a `cast` attached to it, that would also require a BE change perhaps?

1337LutZ:
> I like how this looks, only thing I don't like is the DPS column. I kinda like the full number being shown there instead of it being abbreviated with a K when above a thousand. It has the same length and just makes the information have 2 visual states instead of just 1 (abbreviated and non abbreviated). I think we should revert to just showing the full dps number.

I've un-compacted the DPS/HPS/TDPS values 👍

1337LutZ:
> > @rosenrusinov In the code I saw that Bloodworms and Rune of the Fallen Crusader can't crit their heals, however they can;
> > 
> > * https://classic.warcraftlogs.com/reports/wAtVXFd6YbhmLWya#boss=-2&difficulty=0&wipes=2&type=healing&source=37
> > * https://classic.warcraftlogs.com/reports/wAtVXFd6YbhmLWya#boss=-2&difficulty=0&wipes=2&type=healing&source=16
> > 
> > Can I just add the default crit modifier and `OutcomeHealingCrit` or is this more involved?
> 
> If they work with the standard spell crit when they roll for their critical sure. If not they will require their own implementations of the outcome to calculate the crit. I havent done any testing but from taking a look at the 2 links you shared it doesnt seem to be using the DKs crit chance for them.

Unholy Strength seems to match the player crit in logs and I added it like that for now.

Blood worms range between 2.5-3.5% crit chance, so I think we can either create a variance for this to end up around 3% if needed (this is how it looks in logs, there's not a real correlation between stats or other pets).

proto/api.proto
```diff
@@ -143,6 +143,13 @@ message SimOptions {
 	bool use_labeled_rands = 9; // Use test level RNG.
 }
 
+enum SpellType {
+	SpellTypeCast = 0;
```
NerdEgghead, line 147: The recommended convention for proto enums is to always reserve 0 for "Unknown" (in this case `SpellTypeUnknown`) so that there's always a clear detection for the field not being set in a message (default 0). We've adopted this convention for the bulk of our protos; the only major exception is `Stat` and `PseudoStat`, since those are used to index arrays and we don't want to carry extra array indices around.

sim/core/metrics_aggregator.go
```diff
@@ -119,10 +118,14 @@ type CharacterIterationMetrics struct {
 }
 
 type ActionMetrics struct {
-	IsMelee bool // True if melee action, false if spell action.
+	IsMelee   bool            // True if melee action, false if spell action.
+	IsPassive bool            // True if melee action, false if spell action.
```
NerdEgghead, line 122: Minor note to fix the comment here

sim/core/metrics_aggregator.go
```diff
@@ -308,7 +319,16 @@ func (unitMetrics *UnitMetrics) addSpellMetrics(spell *Spell, actionID ActionID,
 	}
 
 	if !ok {
-		actionMetrics = &ActionMetrics{IsMelee: spell.Flags.Matches(SpellFlagMeleeMetrics)}
+		spellType := proto.SpellType_SpellTypeCast
+		if spell.dots != nil {
```
NerdEgghead, line 323: How are the hit count metrics handled for spells with both an initial and periodic damage component, such as Rake / Thrash / Lacerate for Druids? Currently the spell code for "pure DoT" spells like Rip is set up to manually reduce the `Hits` counter by 1 within `ApplyEffects` so that the correct tick count is shown in the UI, and we don't do this for Rake / Thrash / Lacerate by contrast.

ui/core/components/detailed_results/damage_metrics.tsx
```diff
@@ -0,0 +1,367 @@
...
+								percentage: metric.missPercent,
+							},
+							{
+								name: 'Parry',
+								value: metric.parries,
+								percentage: metric.parryPercent,
+							},
+							{
+								name: 'Dodge',
+								value: metric.dodges,
```
NerdEgghead, line 279: Currently, the displayed Dodge % is completely wrong for bleeds since it gets diluted by the number of ticks. Does this get fixed automatically with your refactor, or are some additional adjustments needed there?

ui/core/proto_utils/names.ts
```diff
@@ -212,10 +224,37 @@ export function getClassStatName(stat: Stat, playerClass: Class): string {
...
+	[SpellSchool.Fire + SpellSchool.Arcane, 'Spellfire'],
+	[SpellSchool.Arcane + SpellSchool.Frost, 'Spellfrost'],
+	[SpellSchool.Frost + SpellSchool.Fire, 'Frostfire'],
+	[SpellSchool.Shadow + SpellSchool.Frost, 'Shadowfrost'],
+]);
+
 export const slotNames: Map<ItemSlot, string> = new Map([
 	[ItemSlot.ItemSlotHead, 'Head'],
 	[ItemSlot.ItemSlotNeck, 'Neck'],
-	[ItemSlot.ItemSlotShoulder, 'Shoulders'],
```
NerdEgghead, line 218: Was this intentional?

ui/core/proto_utils/sim_result.ts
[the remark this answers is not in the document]
1337LutZ: @rosenrusinov I've implemented this in backend using a SpellFlag 👍

1337LutZ:
**Added**
- Glancing blow damage
- Block damage
- Critical Block damage / hits
- Normalized Casts tooltip (Hits + Misses)
- Normalized Missed tooltip

image
image
image

sim/core/metrics_aggregator.go
1337LutZ: I haven't really made changes to how the hits work (as I don't posses the BE knowledge for it), but this is how it looks right now:

**Lacerate**
image

**Rake / Thrash**
image

ui/core/components/detailed_results/damage_metrics.tsx
1337LutZ: No I don't think so, and I'm not sure how to fix this. Wouldn't this require having to track Hits and Ticks (from Dots) separately to actually make it work properly when having a combined spell (with instant + Dot component?)

sim/core/metrics_aggregator.go
1337LutZ: Also see the comment below, I think we need to track these metrics separately and when having a combined spell that has both we can show it properly as well.

proto/api.proto
```diff
@@ -143,6 +143,14 @@ message SimOptions {
 	bool use_labeled_rands = 9; // Use test level RNG.
 }
 
+enum SpellType {
+	SpellTypeUnknown = 0;
+	SpellTypeCast = 1;
+	SpellTypePeriodic = 2;
+	// Needed for Frontend
+	SpellTypeAll = 3;
```
Tharre, line 151: What exactly is the use case here for SpellTypeAll? Combing both doesn't make much sense to me, maybe this could just be  `bool IsPeriodic`?

sim/core/metrics_aggregator.go
```diff
@@ -119,10 +118,14 @@ type CharacterIterationMetrics struct {
...
 type ActionMetrics struct {
-	IsMelee bool // True if melee action, false if spell action.
+	IsMelee   bool            // True if melee action, false if spell action.
+	IsPassive bool            // True if action is applied/cast as a result of another action
+	SpellType proto.SpellType // Spell is cast, periodic (DoT or HoT) or both
 
 	// Metrics for this action, for each possible target.
 	Targets []TargetedActionMetrics
+
+	SpellSchool SpellSchool
```
Tharre, line 128: Minor but move this up to the bool's so we don't waste memory

ui/core/proto_utils/action_id.ts
```diff
@@ -747,6 +749,16 @@ const petNameToActionId: Record<string, ActionId> = {
...
 	'Army of the Dead': ActionId.fromSpellId(42650),
 	Valkyr: ActionId.fromSpellId(71844),
+	Felguard: ActionId.fromSpellId(30146),
+	Felhunter: ActionId.fromSpellId(691),
+	'Ebon Imp': ActionId.fromSpellId(688),
+	Imp: ActionId.fromSpellId(688),
+	Succubus: ActionId.fromSpellId(712),
+	Doomguard: ActionId.fromSpellId(18540),
+	'Summon Doomguard': ActionId.fromSpellId(18540),
+	Infernal: ActionId.fromSpellId(1122),
```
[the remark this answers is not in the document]
1337LutZ: The aim is to do it like WCL does it:
- Have the summoning of the pet as a cast metric on the parent row (including the hit/tick damage averages)
- Have the child rows (abilities) have the actual spell counts and hits etc

ui/core/proto_utils/names.ts
```diff
@@ -212,6 +224,34 @@ export function getClassStatName(stat: Stat, playerClass: Class): string {
...
+export enum SpellSchool {
+	None = 0,
+	Physical = 1 << (ProtoSpellSchool.SpellSchoolPhysical + 1),
+	Arcane = 1 << (ProtoSpellSchool.SpellSchoolArcane + 1),
+	Fire = 1 << (ProtoSpellSchool.SpellSchoolFire + 1),
+	Frost = 1 << (ProtoSpellSchool.SpellSchoolFrost + 1),
+	Holy = 1 << (ProtoSpellSchool.SpellSchoolHoly + 1),
+	Nature = 1 << (ProtoSpellSchool.SpellSchoolNature + 1),
+	Shadow = 1 << (ProtoSpellSchool.SpellSchoolShadow + 1),
+}
```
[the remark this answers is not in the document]
1337LutZ: I did it like this because in the end we should also refactor the schools in backend, since now they are defined in 2 places, 1 in proto for Encounter with incorrect values, and the correct one

proto/api.proto
1337LutZ: I think this is now redundant since I split Hits and Ticks thus merged spells no longer need this

ui/core/proto_utils/action_id.ts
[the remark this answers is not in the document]
Tharre: Huh? WCL doesn't use the summoning spell either, they use the pet name:
image

Obviously we can't mirror that exactly since we don't have a specific pet, but the generic name like `Felhunter` as a replacement makes a lot more sense then `Summon Felhunter`.

sim/core/metrics_aggregator.go
1337LutZ: This is fixed

ui/core/proto_utils/names.ts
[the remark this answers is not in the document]
Tharre: Well, the SpellSchool enum in the proto wasn't meant to encode all spell school combinations in the first place, just the values for the dropdown menu. I do agree that it should just have been an integer instead of the enum, since you really don't gain anything from using the enum; you just created more work for the backend to convert that again. And both should encode to 2 bytes in protobuf.

I guess this can be fixed now with the new versioning thingy if you write a migration for it, though I'm not sure if that's worth it.

But regardless, your code above here will break will break if this gets fixed which is why I suggested hardcoding it.

ui/core/proto_utils/action_id.ts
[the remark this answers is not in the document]
1337LutZ: @Tharre You are right, reverting this. Not sure where I found this 🤔

ui/core/proto_utils/names.ts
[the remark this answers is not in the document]
1337LutZ: Okay I understand. I've hardcoded them.

InDebt:
So let me preface this with: I love this! 😍 

Soo I checked this out and I am kind of lost with MF. It might very well be that there is some stuff that needs to be adjusted core side but right now this looks like this.

grafik

I have three different crit chances shown to me and somehow Average Tick damage is higher than Average Crit damage 🤔 Soo is it calculating as `AverageDamage * %`? At least right now it does not seem intuitive to me.

1337LutZ:
@InDebt I believe I found the issue for the crit %. Mind flay has a CalcOutcome `OutcomeMagicHitAndCrit` which I missed which adds `Hits++`, this should be `OutcomeMagicHitAndCritNoHitCounter` to prevent it act like a instant + Dot spell. Thanks for finding this.

1337LutZ:
@InDebt You were also right about the Crit averages, I was taking the Avg based off the `avgDamage`, I fixed it for all the types of hits in the Damage done breakdown.

InDebt:
Soo the things you fixed work very nice now! There is something else though I noticed.

Not sure why it seems to only effect the Darkmoon Card: Volcano but the damage in the timeline / log is wrong. I guess only there as you say there are not actual test differences. It seems the 'resist' is calculated wrong here. 0% resist is treated as 0% damage done and 10% resist as 10% damage done.

grafik
grafik

I do not see this behaviour in the current live sim, so I guess this is part of theses changes?

InDebt:
Issue has been resolved. Was a migration issue of having the stat refactor branch running before I checked this PR.

kayla-glick (approved):

PR #955, opened 2024-08-21
Add mage T12 bonuses + Fix Fire AOE

kayla-glick:
image


sim/mage/mirror_image_t12.go
```diff
@@ -0,0 +1,106 @@
...
+			// According to the simcraft implementation, the T12 Mirror Images snapshot the owner's crit
+			stats.SpellCritPercent: ownerStats[stats.SpellCritPercent],
+			stats.SpellHitPercent:  ownerStats[stats.SpellHitPercent],
+		}
+	}
+}
+
+func (mi *T12MirrorImage) registerFireballSpell() {
+	mi.Fireball = mi.RegisterSpell(core.SpellConfig{
+		ActionID:     core.ActionID{SpellID: 88082},
```
hillerstorm, line 79: According to simc this should be [99062](https://www.wowhead.com/cata/spell=99062/fireball) (which also fits the damage range). [88082](https://www.wowhead.com/cata/spell=88082/fireball#used-by-npc) seems to be used by a different [Mirror Image](https://www.wowhead.com/cata/npc=47244/mirror-image) than the T12 summons: [triggered spell](https://www.wowhead.com/cata/spell=99063/mirror-image#triggered-by), [summons this](https://www.wowhead.com/cata/npc=53438/mirror-image).

sim/mage/items.go
```diff
@@ -28,33 +30,88 @@ var ItemSetFirelordsVestments = core.NewItemSet(core.ItemSet{
...
+					mage.hotStreakProcChance += 0.30
+				},
+				OnExpire: func(aura *core.Aura, sim *core.Simulation) {
+					mage.brainFreezeProcChance -= .15
+					mage.hotStreakProcChance -= .30
+				},
+			}))
+
+			if mage.arcanePowerCostMod != nil {
+				mage.arcanePowerCostMod.UpdateFloatValue(-1)
```
hillerstorm, line 70: 🤔 shouldn't this be `-0.2`? The default value seems to be `0.1`, aka 10% increased cost and the 4pc should change it to `-0.1`. This line makes it free instead of having a 10% reduced cost 😄

sim/mage/items.go
```diff
@@ -28,33 +30,88 @@ var ItemSetFirelordsVestments = core.NewItemSet(core.ItemSet{
...
+			}
+		},
+	},
+})
+
+// T13
+var ItemSetTimeLordsRegalia = core.NewItemSet(core.ItemSet{
```
hillerstorm, line 77: Ment to add this in a different pr? Seems to be T12 😃

sim/mage/mirror_image_t12.go
kayla-glick: Oh nice catch!

sim/mage/items.go
kayla-glick: Oh maybe! I'll look at what that method does

sim/mage/items.go
kayla-glick: Yeah I was going to start T13 then I thought I should release T12 first 😄

sim/core/dot.go
```diff
@@ -154,9 +156,12 @@ func (dot *Dot) CopyDotAndApply(sim *Simulation, originaldot *Dot) {
 	dot.Activate(sim)
 
 	// recreate with new period, resetting the next tick.
+	if dot.tickAction != nil {
+		dot.tickAction.Cancel(sim)
+	}
```
kayla-glick, line 161: This frequently breaks if you don't try to cancel an existing action

sim/mage/talents_fire.go
```diff
@@ -123,9 +123,10 @@ func (mage *Mage) applyHotStreak() {
 
 	ImprovedHotStreakProcChance := float64(mage.Talents.ImprovedHotStreak) * 0.5
 
-	// TODO: Is this supposed to be fully buffed Spell Crit % or only from gear Rating?
+	// Simcraft uses a reference from ElitistJerks that's no longer available, but their formula is
+	// max(0, -2.73 * player crit + 0.95)
 	baseCritPercent := mage.GetStat(stats.SpellCritPercent) + mage.GetStat(stats.CritRating)/core.CritRatingPerCritPercent
-	BaseHotStreakProcChance := float64(-2.7*baseCritPercent/100 + 0.9) // EJ settled on -2.7*critChance+0.9
+	mage.hotStreakProcChance = max(0, -2.73*baseCritPercent/100+0.95)
```
kayla-glick, line 129: Updated the Hot Streak calculation with values from simcraft

1337LutZ: You can probably link it to this post: https://web.archive.org/web/20120208064232/http://elitistjerks.com/f75/t110326-cataclysm_fire_mage_compendium/p6/#post1831143

sim/mage/items.go
```diff
@@ -28,33 +30,61 @@ var ItemSetFirelordsVestments = core.NewItemSet(core.ItemSet{
...
+					mage.brainFreezeProcChance += .15
+					mage.hotStreakProcChance += 0.30
+				},
+				OnExpire: func(aura *core.Aura, sim *core.Simulation) {
+					mage.brainFreezeProcChance -= .15
+					mage.hotStreakProcChance -= .30
+				},
+			}))
+
+			if mage.arcanePowerCostMod != nil {
```
InDebt, line 69: Did you debug / verify this? I think I had to do a callback here because ItemEffects were initialized before the talents. Only the agent constructor is called before.

```go
mage.OnSpellRegistered(func(spell *core.Spell) {
  if (spell.ClassSpellMask != MageSpellArcanePower) return;

  // do stuff
}
```

BTW while checking this I noticed that the ArcanePower spell doesn't even define a class mask, thus, the `ArcaneFlows` talent isn't applied to it properly :D

kayla-glick: Oh I forgot to check that thank you. And oof yeah it looks like there's a class mask defined but not applied to the AP spell

PR #964, opened 2024-08-22
Fix Variable Pulse Lightning Capacitor

### LINKED ISSUE #941: [Item] Variable Pulse Lightning Capacitor currently only procs from direct abilities
**Describe the bug**
Currently the trinket only procs from direct spells, both casts and hits, but not periodic effects, causing it to appear extremely weak on the sims.

**To Reproduce**
1. Sim

**Expected behavior**
It is supposed to both stack up and also proc from dots.

**Proof**
<https://www.youtube.com/watch?v=HJV2yHDk_r0> - Throughout this entire video you can see the person both gain and expunge the Lightning Charges to the boss while not doing anything / just channeling Mind Flay, so it is definitely dots that are causing this.

https://www.wowhead.com/cata/item=68925/variable-pulse-lightning-capacitor#comments - Many comments in here confirming the same things:

"1. the proc is triggered by your dmg spell - dots or direct"
"Just got this on my priest, went straight to the training dummies and can confirm that it procs from DoT ticks."
"It can proc from dot crits and even abilities like flame orb (which aren't a direct hit OR a dot) trigger it"

hillerstorm:
Fixes #941

### DIFF
--- a/sim/common/cata/damage_procs.go
+++ b/sim/common/cata/damage_procs.go
@@ -70,7 +70,7 @@ func init() {
 		core.MakePermanent(core.MakeProcTriggerAura(&character.Unit, core.ProcTrigger{
 			Name:       "Electrical Charge Aura",
 			ActionID:   core.ActionID{ItemID: 68925},
-			Callback:   core.CallbackOnSpellHitDealt,
+			Callback:   core.CallbackOnSpellHitDealt | core.CallbackOnPeriodicDamageDealt,
 			ProcMask:   core.ProcMaskSpellOrProc,
 			ProcChance: 1,
 			Outcome:    core.OutcomeCrit,
@@ -115,7 +115,7 @@ func init() {
 		core.MakePermanent(core.MakeProcTriggerAura(&character.Unit, core.ProcTrigger{
 			Name:       "Electrical Charge Aura",
 			ActionID:   core.ActionID{ItemID: 69110},
-			Callback:   core.CallbackOnSpellHitDealt,
+			Callback:   core.CallbackOnSpellHitDealt | core.CallbackOnPeriodicDamageDealt,
 			ProcMask:   core.ProcMaskSpellOrProc,
 			ProcChance: 1,
 			Outcome:    core.OutcomeCrit,
--- a/sim/death_knight/blood/TestBlood.results
+++ b/sim/death_knight/blood/TestBlood.results
@@ -1454,16 +1454,16 @@ dps_results: {
 dps_results: {
  key: "TestBlood-AllItems-VariablePulseLightningCapacitor-68925"
  value: {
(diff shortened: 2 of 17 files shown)


sim/common/cata/damage_procs.go
```diff
@@ -70,7 +70,7 @@ func init() {
 		core.MakePermanent(core.MakeProcTriggerAura(&character.Unit, core.ProcTrigger{
 			Name:       "Electrical Charge Aura",
 			ActionID:   core.ActionID{ItemID: 68925},
-			Callback:   core.CallbackOnSpellHitDealt,
+			Callback:   core.CallbackOnSpellHitDealt | core.CallbackOnPeriodicHealDealt,
```
zoristaken, line 73: this is meant to be periodic damage right

hillerstorm: Haha that's embarrassing... Thanks autocomplete 😅

sim/common/cata/damage_procs.go
```diff
@@ -70,7 +70,7 @@ func init() {
...
 			ProcMask:   core.ProcMaskSpellOrProc,
```
zoristaken, line 74: > also not sure if I used the correct mask here, I checked Bell of Enraging Resonance and they have the same proc mask on the db, but that one was implemented with core.ProcMaskSpellDamage,

same here

sim/common/cata/damage_procs.go
```diff
@@ -115,7 +115,7 @@ func init() {
 		core.MakePermanent(core.MakeProcTriggerAura(&character.Unit, core.ProcTrigger{
 			Name:       "Electrical Charge Aura",
 			ActionID:   core.ActionID{ItemID: 69110},
-			Callback:   core.CallbackOnSpellHitDealt,
+			Callback:   core.CallbackOnSpellHitDealt | core.CallbackOnPeriodicDamageDealt,
 			ProcMask:   core.ProcMaskSpellOrProc,
```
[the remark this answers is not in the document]
hillerstorm: [this one](https://www.wowhead.com/cata/spell=96887/variable-pulse-lightning-capacitor) does have `Can Proc From Procs` which usually means `ProcMaskProc`. I think we have the incorrect one for Bell actually, not the other way around 🤔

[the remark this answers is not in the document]
hillerstorm: I fixed this for all the strength trinkets a few weeks ago but I think someone should have a look at all cata trinkets and check for that proc mask.

InDebt (approved):
Shadow changes are fine. Although I'm uncertain right now what kind of crit feedback loop we have that we loose 40 dps in the tests even thoguh the Crit -> Hit change is for a 0 dps hit :D

PR #975, opened 2024-08-25
[UI] Preset configurations

1337LutZ:
- Added the ability to add preset configurations that consist of:
  - Gear
  - Talents
  - Rotation
  - EP
  - Encounter (& Healing Model) - This has support for `Export > Link` imports for convenience.
- Added Preset Configuration picker that will show on each tab that is eligible

**Examples**
**DK**
image
**Guardian Druid**
image
Doesn't show here because Gear isn't effected by any of the presets.
image


ui/core/encounter.ts
```diff
@@ -198,6 +196,9 @@ export class Encounter {
 	}
 
 	static updateProtoVersion(proto: EncounterProto) {
+		if (!(proto.apiVersion < CURRENT_API_VERSION)) {
```
1337LutZ, line 199: @NerdEgghead Moved the logic here to not have to duplicate the update logic everywhere you want to apply it with the same if statement.

ui/core/proto_utils/stats.ts
1337LutZ, line 1: Everything here except moving `updateProtoVersion` check is linting/formatting

NerdEgghead (approved):

PR #981, opened 2024-08-27
Fix target input labels not changing when swapping presets

NerdEgghead:
On branch feature/preset-configurations
 Changes to be committed:
	modified:   ui/core/components/encounter_picker.ts
	modified:   ui/core/components/input.tsx


NerdEgghead:
@1337LutZ I think this fixes the target inputs bug I mentioned to you, but would appreciate another pair of eyes on it to make sure I didn't do anything stupid.

1337LutZ (approved):

PR #987, opened 2024-08-29
[UI] Remove rating from reforge interface

1337LutZ:
With the stat refactor the "rating" value became even more confusing as it now only points to gear rating.
To remove some of the confusion I've removed the Rating fields (as they resemble total rating which is just a derivative of %)

image
image

### DIFF
--- a/ui/core/components/suggest_reforges_action.tsx
+++ b/ui/core/components/suggest_reforges_action.tsx
@@ -9,7 +9,7 @@ import { Player } from '../player';
 import { Class, ItemSlot, PseudoStat, Spec, Stat } from '../proto/common';
 import { StatCapType } from '../proto/ui';
 import { Gear } from '../proto_utils/gear';
-import { shortSecondaryStatNames, slotNames, statCapTypeNames } from '../proto_utils/names';
+import { slotNames, statCapTypeNames } from '../proto_utils/names';
 import { pseudoStatIsCapped, StatCap, Stats, UnitStat } from '../proto_utils/stats';
 import { SpecTalents } from '../proto_utils/utils';
 import { Sim } from '../sim';
@@ -41,16 +41,12 @@ type StatTooltipContent = { [key in Stat]?: () => Element | string };
 const STAT_TOOLTIPS: StatTooltipContent = {
 	[Stat.StatMasteryRating]: () => (
 		<>
-			Rating: <strong>excluding</strong> your base mastery
-			<br />
-			%: <strong>including</strong> your base mastery
+			<strong>Including</strong> your base mastery
 		</>
 	),
 	[Stat.StatHasteRating]: () => (
 		<>
-			Rating: final rating <strong>including</strong> all buffs/gear.
-			<br />
-			%: final percentage value <strong>including</strong> all buffs/gear.
+			Final percentage value <strong>including</strong> all buffs/gear.
 		</>
 	),
 };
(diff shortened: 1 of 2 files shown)


ui/core/components/pickers/number_picker.ts
```diff
@@ -8,6 +9,7 @@ export interface NumberPickerConfig<ModObject> extends InputConfig<ModObject, nu
 	id: string;
 	// Whether the picker represents a float value. Default `false`
 	float?: boolean;
+	fractionCount?: number;
```
NerdEgghead, line 12: Minor nit, but this name confused me until I saw what it meant. Renaming to `maxDecimalDigits` or `decimalPrecision` would be clearer imo.

ui/mage/fire/presets.ts
NerdEgghead, line 1: Do you need to make this change for Shadow Priest presets too? Might be other specs to double check as well, not sure.

ui/core/components/suggest_reforges_action.tsx
```diff
@@ -525,24 +509,26 @@ export class ReforgeOptimizer {
...
 												name: `${key} - ${percentValue.toFixed(2)}%`,
-												value: ratingValue,
+												value: percentOrPointsValue,
 											};
 										}),
 									].sort((a, b) => a.value - b.value),
 									enableWhen: () => this.isAllowedToOverrideStatCaps || !this.softCapsConfig.some(config => config.unitStat.equals(unitStat)),
 									getValue: () => {
-										let ratingValue = unitStat.convertDefaultUnitsToRating(this.statCaps.getUnitStat(unitStat))!;
+										let ratingValue = this.statCaps.getUnitStat(unitStat);
```
NerdEgghead, line 525: I don't think this is correct, it will break for Expertise. You need to call `convertDefaultUnitsToPercent()` so that it will handle unit conversions automatically. Also rename the variable to `percentValue` because it's incorrect and very confusing currently.

ui/core/components/suggest_reforges_action.tsx
```diff
@@ -525,24 +509,26 @@ export class ReforgeOptimizer {
...
 										if (unitStat.equalsStat(Stat.StatMasteryRating)) ratingValue = this.toVisualBaseMasteryRating(ratingValue);
 
 										return ratingValue;
 									},
 									setValue: (_eventID, _player, newValue) => {
-										const statValue = unitStat.equalsStat(Stat.StatMasteryRating) ? (newValue + this.baseMastery) : unitStat.convertRatingToDefaultUnits(newValue)!;
+										const statValue = unitStat.equalsStat(Stat.StatMasteryRating) ? newValue + this.baseMastery : newValue;
```
NerdEgghead, line 531: Same thing here, you need to copy the logic from the `percentagePicker` so it will work for both Stats and PseudoStats and automatically convert units.

ui/core/components/suggest_reforges_action.tsx
```diff
@@ -525,24 +509,26 @@ export class ReforgeOptimizer {
 									values: [
 										{ name: 'Select preset', value: 0 },
 										...[...statPresets.keys()].map(key => {
-											const ratingValue = statPresets.get(key)!;
-											const percentOrPointsValue = unitStat.convertRatingToPercent(ratingValue)!;
-											const percentValue = unitStat.equalsStat(Stat.StatMasteryRating) ? (percentOrPointsValue * this.player.getMasteryPerPointModifier()) : percentOrPointsValue;
+											const percentOrPointsValue = statPresets.get(key)!;
```
NerdEgghead, line 512: It's a bit confusing that the keys for `statPresets` are proper Rating stats, but the values are percentages rather than Rating equivalents. Maybe that's okay, but there should some documentation clarifying that the expected units for the values are percent.

ui/core/components/suggest_reforges_action.tsx
```diff
@@ -525,24 +509,26 @@ export class ReforgeOptimizer {
...
+											const percentValue = unitStat.equalsStat(Stat.StatMasteryRating)
+												? percentOrPointsValue * this.player.getMasteryPerPointModifier()
+												: percentOrPointsValue;
+
 											return {
 												name: `${key} - ${percentValue.toFixed(2)}%`,
-												value: ratingValue,
+												value: percentOrPointsValue,
```
NerdEgghead, line 519: should be `percentValue` rather than `percentOrPointsValue` here

ui/core/components/suggest_reforges_action.tsx
```diff
@@ -525,24 +509,26 @@ export class ReforgeOptimizer {
...
 											};
 										}),
 									].sort((a, b) => a.value - b.value),
 									enableWhen: () => this.isAllowedToOverrideStatCaps || !this.softCapsConfig.some(config => config.unitStat.equals(unitStat)),
 									getValue: () => {
-										let ratingValue = unitStat.convertDefaultUnitsToRating(this.statCaps.getUnitStat(unitStat))!;
+										let ratingValue = this.statCaps.getUnitStat(unitStat);
 										if (unitStat.equalsStat(Stat.StatMasteryRating)) ratingValue = this.toVisualBaseMasteryRating(ratingValue);
```
NerdEgghead, line 526: This logic needs to match the percentage picker, not the (now deprecated) rating picker.

ui/core/components/suggest_reforges_action.tsx
NerdEgghead: In fact, if possible, try to re-use all the same code in `percentagePicker` for the presets picker rather than copy-pasting everything.

NerdEgghead (changes requested):

ui/core/components/suggest_reforges_action.tsx
```diff
@@ -525,24 +509,26 @@ export class ReforgeOptimizer {
 									values: [
 										{ name: 'Select preset', value: 0 },
 										...[...statPresets.keys()].map(key => {
-											const ratingValue = statPresets.get(key)!;
-											const percentOrPointsValue = unitStat.convertRatingToPercent(ratingValue)!;
-											const percentValue = unitStat.equalsStat(Stat.StatMasteryRating) ? (percentOrPointsValue * this.player.getMasteryPerPointModifier()) : percentOrPointsValue;
+											const percentOrPointsValue = statPresets.get(key)!;
+											const percentValue = unitStat.equalsStat(Stat.StatMasteryRating)
+												? percentOrPointsValue * this.player.getMasteryPerPointModifier()
+												: percentOrPointsValue;
```
NerdEgghead, line 515: Oh also, if the convention is going to be that the units for `statPresets` values will always be percent, then this conversion should not be done, and it should be assumed that the `percentValue` is directly what you are reading from the map.

ui/core/components/suggest_reforges_action.tsx
NerdEgghead: Here's an idea for a more elegant way to handle this and keep the units always consistent with the key:
- Instead of defining `statSelectionPresets` as `Map<Stat, Map<string, number>>`, instead define it as `Array<UnitStatPresets>`, where `UnitStatPresets` is an interface with `unitStat: UnitStat` and `presets: Map<string, number>` as required fields.
- Then instead of checking `this.statSelectionPresets?.get(rootStat)`, we do `this.statSelectionPresets?.find((entry) => entry.unitStat.equals(unitStat))`.
- Now we don't have to prescribe some new convention for the units, and keep the global convention I've adopted across the front-end code where the assumed units are always given in the name of the stat itself. So `SpellHastePercent` presets would be assumed to be given in percent units, while `MasteryRating` presets would be assumed to be given in Rating units.
- Then you can just call `convertDefaultUnitsToPercent` when building the display strings to show in percent, and unit conversions will happen automatically.

It's slightly more code than what you have currently, but preserves the integrity of units always being consistent with the name of the stat, which is something I really want to enforce after spending so much time on the stats refactor.

ui/mage/fire/presets.ts
1337LutZ: Oh damn you are right, totally forgot I did this for SP. I double checked and it's only;
- Fire
- Shadow
- Warlocks

ui/core/components/suggest_reforges_action.tsx
1337LutZ: @NerdEgghead I liked that idea. I've made the changes to all the "presets", is this what you had in mind?

NerdEgghead (approved):
Looks much cleaner now, thanks!

Issue #946, opened 2024-08-20
[Item] Dragonwrath, Tarecgosa's Rest


InDebt:
> [!IMPORTANT]
> With the approach of the next phase, the desire to accuratley simulate and gauge `Dragonwrath, Tarecgosa's Rest` will be higher and higher. This Staff is the newest legendary introduced with the Firelands raid and will be in high demand for every caster. Therefore, it's very desirebly to provide a high quality implementation for this item, as it has many class specific interactions. This issue shall be used to gather all information we can, as it's unlikely the staff will be available for testing on the upcoming PTR. Please contribute any old resources and references you have. Take the results of the simulations with the staff with bit of scepticism until we have larger testing and reference datasets.

> [!NOTE]
> This issue will be constantly updated with the latest references and feedback gathered

## General behaviour
* When equiped the staff provides the ability to duplicate the casters damaging spells.
* Direct damaging spells will be replicated as a direct copy of that same spell
* Periodic damaging ticks will be replicated causing the exact same amount of damage as arcane damage through [Wrath of Tarecgosa](https://www.wowhead.com/cata/spell=101085/wrath-of-tarecgosa) [^1]
  * As it's another spell, players will not benefit from talents here. A good exaple would be `Shadow Orbs` which can trigger from `Shadow Word: Pain` ticks. These will not be triggered by a tick replication from the staff.
* The replicated spells from direct casts behave the same way a cast would that player's cast themselves. i.E. consuming procs and benefitting from the same modifiers. [^1]
* Spells are replicated on damage. Causing spells with projectiles to be replicated on impact [^4]
* Even though it is stated that the Staff has an internal cooldown, it seems to be around the 10ms visible through wowhead. [^13]
  * Procs occuring immediatly after another can be seen here [^12] 56s

## Class specifics
* There seem to be different proc chances for different classes as indicated by blizzard.
> Furthermore, the proc chance is adjusted favorably for pet specs that derive a significant portion of their DPS from pets.
[^2]

### Proc Chances per Spec
| Class | Spec | Proc Chance | Reference |
| ----- | ----- | :-----------: | ---------- |
| Druid | Balance | 8% | [^3] [^10] [^19] |
| Mage | Arcane | 9.16% | simc (4.3) |
| Mage | Fire | 8.0% | [^16] |
| Mage | Frost | 9.16% | simc (4.3) |
| Priest | Discipline | ? | - |
| Priest | Shadow | 9% | [^17] |
[...]

PR #989, opened 2024-08-30
[Item] Dragonwrath, Tarecgosa's Rest

InDebt:
This is the implementation PR for Dragonwrath, Tarecgosa's Rest.
Implements #946 for further information.

> [!NOTE]
> This is an early draft encouring early feedback and participation. Class specific implementations are highly encouraged as I don't know how much time I have a week to get this finished.

## Current status
* [x] Implement baseline handling
  * [x] Proper (Fluent?) API missing to add custom handlers from class folders
* [x] Add Shadow support
* [ ] Add Discipline support
* [x] Add Elemental support
* [x] Add Balance support
* [x] Add Affliction support
* [x] Add Demonology support
* [x] Add Destruction support
* [x] Add Arcane Mage support 
* [x] Add Fire Mage support 
* [x] Add UI disclaimers

## Open topics
On retail Spells instantiate casts which carry a lot of meta information on the actual circumstances and values used during their instantiation. Like base spell power when cast, attackpower and so on. Most importantly casts know the meta data for their calculations. I.e. The pulse lightning capaciatotor knows the charges it was triggered with or Fulmination the lightning shield stacks it was triggered with. In theory DRT copies casts not spells. Therfore the meta data is the same and the same calculations will be used causing fulimnation and capacitator to use the same aura charges for damage calculations. This is something rather tricky to realize right now in our backend and we have to see if all instances can simply "missuse" the `BonusSpellDamage` field for such things.


sim/common/cata/dragonwrath.go
```diff
@@ -0,0 +1,280 @@
...
+				if config == nil {
+					// TODO: HANDLE BETTER
+					panic("DTR not supported for this spec yet")
+				}
+
+				if val, ok := config.spellConfig[spell.SpellID]; ok && val&supressImpact > 0 {
+					return
+				}
+
+				// make sure the same spell impact can only trigger once per timestamp (AoE Impact spells like Arcane Explosion or Mind Sear)
```
Polynomix, line 182: This might prevent multiple Chain Lightning Overloads from being duplicated

InDebt: You might be right. Trigger/Proc systems are always the hardest to get right even when I was working on p-servers as they heavily depend on how blizz actually implemented stuff.

I suppose we have the following:
* `TARGET_UNIT_TARGET_ENEMY` - proc per target even chain targets
* `TARGET_UNIT_DEST_AREA_ENEMY` and `TARGET_UNIT_SRC_AREA_ENEMY` - per cast no matter how many targets
  * Usually those are spells like  concecrate, blizzard, arcane explosion or mind sear.

Obviously we don't have access to the target enums for the spells but I guess we need to be able to configure those spells in the class setup. I'll take a look to provide a clean API.

InDebt: I more and more wish we had a differenitation between spells and casts i guess xD

1337LutZ:
> @1337LutZ Do you have the time and would want to look into how we could display a, perferably spec specific disclaimer for DTR in the UI? Right now I have desigend the staff's implementation that allows us to implement it spec by spec. So the accuracy may very from spec to spec.

Sure thing!
Would this just be coded in the the `sim.ts` config? (I assume yes)
I'll have time this weekend to look into this.

sim/druid/balance/dragonwrath.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	"github.com/wowsims/cata/sim/common/cata"
+	"github.com/wowsims/cata/sim/core/proto"
+)
+
+func init() {
+	cata.CreateDTRClassConfig(proto.Spec_SpecBalanceDruid, 0.072).
+		AddSpell(8921, cata.NewDragonwrathSpellConfig().SupressImpact()). // Moonfire
+		AddSpell(93402, cata.NewDragonwrathSpellConfig().SupressImpact()) // Sunfire
```
stn-ms, line 11: Please consider having a spell proc chance override. It seems like there are spells with very low proc chance, like Mushrooms and Typhoon - https://www.mmo-champion.com/threads/967211-FAQ-Dragonwrath-Tarecgosa-s-Rest-and-you-a-Balance-Druid

InDebt: Right now I  assume the 'low' proc chance reported is due to the fact that both typhoon and shrooms only get a single proc rool per 'cast' which hits serveral targets, which is to be expected.

Adding a custom proc chance later on would be very easy. I can just add a proc chance overwrite on the `SpellConfig` for the class if testing should confirm this.

1337LutZ:
Just merged master into this branch since there were quite some conflicts

1337LutZ:
@InDebt Would it be an idea to for now only enable DTR under the Experimental flag?
That way it allows "hardcore" Sim users to play around with it, whilst the "click-and-go" users will see it as a stat stick when comparing to other classes. I believe otherwise all Class Discords will get a bunch of questions on why spec XXX does so much more DPS.

sim/mage/blast_wave.go
```diff
@@ -43,59 +43,57 @@ func (mage *Mage) registerBlastWaveSpell() {
 				baseDamage *= sim.Encounter.AOECapMultiplier()
 				spell.CalcAndDealDamage(sim, aoeTarget, baseDamage, spell.OutcomeMagicHitAndCrit)
 			}
-			if targetCount > 1 {
-				mage.FlamestrikeBW.Cast(sim, target)
-			}
+			// if targetCount > 1 {
```
1337LutZ, line 46: Should this commented code be removed?

InDebt: Yes. I moved the Improved Flamestrike logic into the talent and commented the previous code during testing. Will remove it, as it seems to work as I want.

InDebt:
> @InDebt Would it be an idea to for now only enable DTR under the Experimental flag? That way it allows "hardcore" Sim users to play around with it, whilst the "click-and-go" users will see it as a stat stick when comparing to other classes. I believe otherwise all Class Discords will get a bunch of questions on "Why spec XXX does so much more DPS, this is unfair".

Right now damage gains are close together. Do we even get the 'experimental' flag with the proto request so I could disable / enable class support on the backend? Because otherwise we'd need to thide the staff alltogether behind that flag.

1337LutZ:
> > @InDebt Would it be an idea to for now only enable DTR under the Experimental flag? That way it allows "hardcore" Sim users to play around with it, whilst the "click-and-go" users will see it as a stat stick when comparing to other classes. I believe otherwise all Class Discords will get a bunch of questions on "Why spec XXX does so much more DPS, this is unfair".
> 
> Right now damage gains are close together. Do we even get the 'experimental' flag with the proto request so I could disable / enable class support on the backend? Because otherwise we'd need to thide the staff alltogether behind that flag.

As discussed in Discord let's wait with merging for now 👌

InDebt:
@1337LutZ and others of course. I think we're in a state where we could give this out for preliminary testing.
I've updated the discliamer for the weapon. Checked I think all the reported interactions and adjusted accordingly after the PTR.

There will certainly be minor things I will inevitably have missed, so I think it's a good point to let others help out find those things 😁 .

Lutz already went and did some math and at least the current gains from DTR seem to be within expectations.

1337LutZ:
> @1337LutZ and others of course. I think we're in a state where we could give this out for preliminary testing. I've updated the discliamer for the weapon. Checked I think all the reported interactions and adjusted accordingly after the PTR.
> 
> There will certainly be minor things I will inevitably have missed, so I think it's a good point to let others help out find those things 😁 .
> 
> Lutz already went and did some math and at least the current gains from DTR seem to be within expectations.

Super awesome, and thanks for all your hard work on making a extensible implementation. ❤️ 
I've shared some Local builds with DTR implemented with some of the Discord big brains and most see it in line and have already provided feedback for some changes that I relayed to you. 👍

1337LutZ (approved):

stn-ms (approved):

PR #996, opened 2024-09-04
[Rogue] Updates for T12

TheBackstabi:
Adds T12 presets for all three rogue specs.

Includes some fixes for various talents that were improperly implemented.


ui/rogue/subtlety/presets.ts
```diff
@@ -6,34 +6,36 @@ import { SavedTalents } from '../../core/proto/ui';
 import { Stats } from '../../core/proto_utils/stats';
 import SubtletyApl from './apls/subtlety.apl.json';
 import P1SubtletyGear from './gear_sets/p1_subtlety.gear.json';
+import P2SubtletyGear from './gear_sets/p2_subtlety.gear.json';
```
1337LutZ, line 9: We should probably follow Blizzard Phases, so this should be Phase 3. Other specs have already called it P3 as well.

1337LutZ (approved):

TheBackstabi:
I don't have merge perms in case that's been the holdup 😅

1337LutZ:
> I don't have merge perms in case that's been the holdup 😅

Ahhh sorry I didn't know 😭😂

PR #1006, opened 2024-09-12
[Balance] Preset work, T12, T13 and cleanup

stn-ms:


stn-ms:
@rosenrusinov @InDebt @1337LutZ I removed some options from the proto file that are no longer needed and had to bump the version, according to the log. Do I actually need to do something in updateProtoVersion() about those?

NerdEgghead:
> @rosenrusinov @InDebt @1337LutZ I removed some options from the proto file that are no longer needed and had to bump the version, according to the log. Do I actually need to do something in updateProtoVersion() about those?

It depends on whether the fields you removed are saved in user browser data, and require special handling in the UI for migrating the old data so that sim behavior remains consistent. If this is the case, then you should add an `apiVersion` field to the affected proto, and add front-end code that will process an out-of-date proto.

On the other hand, if sims will run as expected with these old fields simply ignored by the back-end, then there is no reason to delete the unused fields from the proto and increment the version number. Instead, you can simply flag the fields as deprecated and silently ignore them, such as in [this](https://github.com/wowsims/cata/pull/979/commits/2b0c6d07564f474646cbb52ea56f44ccf3426b77) example.

stn-ms:
> > @rosenrusinov @InDebt @1337LutZ I removed some options from the proto file that are no longer needed and had to bump the version, according to the log. Do I actually need to do something in updateProtoVersion() about those?
> 
> It depends on whether the fields you removed are saved in user browser data, and require special handling in the UI for migrating the old data so that sim behavior remains consistent. If this is the case, then you should add an `apiVersion` field to the affected proto, and add front-end code that will process an out-of-date proto.
> 
> On the other hand, if sims will run as expected with these old fields simply ignored by the back-end, then there is no reason to delete the unused fields from the proto and increment the version number. Instead, you can simply flag the fields as deprecated and silently ignore them, such as in [this](https://github.com/wowsims/cata/pull/979/commits/2b0c6d07564f474646cbb52ea56f44ccf3426b77) example.

Thanks @NerdEgghead, that's what I did now.

sim/druid/items.go
```diff
@@ -163,6 +163,39 @@ var ItemSetObsidianArborweaveRegalia = core.NewItemSet(core.ItemSet{
...
+	},
+})
+
+// T13 Balance
+var ItemSetDeepEarthRegalia = core.NewItemSet(core.ItemSet{
+	Name: "Deep Earth Regalia",
+	Bonuses: map[int32]core.ApplyEffect{
+		// Insect Swarm increases all damage done by your Starfire, Starsurge, and Wrath spells against that target by 3%
+		2: func(agent core.Agent) {
+			//druid := agent.(DruidAgent).GetDruid()
```
1337LutZ, line 182: Can this commented code be removed?

ui/druid/balance/sim.ts
```diff
@@ -106,16 +104,16 @@ const SPEC_CONFIG = registerSpecConfig(Spec.SpecBalanceDruid, {
...
-		epWeights: [Presets.P1_EP_PRESET],
+		epWeights: [Presets.StandardEPWeights],
 		// Preset talents that the user can quickly select.
 		talents: [Presets.StandardTalents],
-		rotations: [Presets.PresetRotationDefault],
+		rotations: [Presets.T11PresetRotation, Presets.T12PresetRotation],
 		// Preset gear configurations that the user can quickly select.
-		gear: [Presets.PreraidPresetGear, Presets.T11PresetGear, Presets.T12PresetGear],
+		gear: [Presets.PreraidPresetGear, Presets.T11PresetGear, Presets.T12PresetGear, Presets.T13PresetGear],
 	},
```
1337LutZ, line 113: You could also add a `builds` preset, which will allow for gear + rotation swapping in 1 click. You can check Feral or Fury for examples.

1337LutZ (changes requested):

sim/druid/items.go
stn-ms: done

ui/druid/balance/sim.ts
stn-ms: also done. i'm not sure why my first preset button doesn't get 'activated' though, might be some cache that's being weird locally..

1337LutZ: Ah, I was talking about the `builds` array which allows you to swap Gear + Rotation + EP Weights etc with 1 button press

1337LutZ (approved):

PR #1009, opened 2024-09-13
[FIX] Adjust VPLC proc rate and add warning

1337LutZ:
- Adjusted proc rate to 50% based off @InDebt research
- Added proc rate warning
image

### DIFF
--- a/sim/common/cata/damage_procs.go
+++ b/sim/common/cata/damage_procs.go
@@ -44,7 +44,8 @@ func init() {
 
 	core.NewItemEffect(68925, func(agent core.Agent) {
 		character := agent.GetCharacter()
-		procChance := 0.15 //TODO: simc proc chance, verify when available
+		// Proc chance determined to be p=.48 by video research - Researched by InDebt & Frostbitten
+		procChance := 0.5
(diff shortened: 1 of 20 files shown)


1337LutZ:
InDebt will add his research notes here later

1337LutZ:
Post talking about ~50% proc rate & never exceeding ~3 stacks:
- https://www.mmo-champion.com/threads/971361-Stealth-Buff-Variable-Pulse-Lightning-Capacitor?p=13067475&viewfull=1#post13067475
- https://www.mmo-champion.com/threads/971361-Stealth-Buff-Variable-Pulse-Lightning-Capacitor?p=13069005&viewfull=1#post13069005

InDebt:
So Frostbitten was nice enough to count some procs and at how many stacks they occurred. See graphic below
image

The underlying distribution should be a geometric distribution. This we can calculate the average and estimate the underlying propability with $\frac{1}{\mu}$
$\mu = 2.14583\bar{3}$
$p=\frac{1}{2.14583\bar{3}} =0.48$
Based on the small sample size 0.5 is a good estimation for the underlying propability

PR #1010, opened 2024-09-13
[DEMONOLOGY][UI] Add post prepull snapshot mana

1337LutZ:
- Allows to adjust mana after swapping from prepull set
- Added UI for post prepull mana
- Added OtherID for Prepull specific actions

image
image
image

### DIFF
--- a/proto/common.proto
+++ b/proto/common.proto
@@ -956,6 +956,7 @@ enum OtherAction {
 	OtherActionSolarEnergyGain = 18; // For balance druid solar energy
 	OtherActionLunarEnergyGain = 19; // For balance druid lunar energy
 	OtherActionMove = 20; // Used by movement to be able to show it in timeline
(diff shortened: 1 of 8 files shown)


Tharre:
Adding a whole new category of mana metric just for that seems overkill, but otherwise LGTM 😄

1337LutZ:
> Adding a whole new category of mana metric just for that seems overkill, but otherwise LGTM 😄

I was unsure how to access the available one as I'm not that experienced with the Backend code. Perhaps you can show me? 👌

Tharre:
> I was unsure how to access the available one as I'm not that experienced with the Backend code. Perhaps you can show me? 👌

There's none, I meant I wouldn't have added `OtherAction_OtherActionPrepull` specifically for this, I'd have just used the action id of meta for it. But your solution works fine.

PR #1021, opened 2024-09-17
Unify Ignite code and add optional munching model

NerdEgghead:
While working on a Feral 2pT12 implementation, I realized that we now have 3 nearly identical copy-pasted implementations of Ignite in the codebase, which is something we should seek to avoid whenever possible. This PR implements a shared utility for modeling any Ignite-style effect, and then calls that utility with a few lines of boilerplate for modeling both actual Mage Ignite and the Rogue/Ret/Feral 2pT12 bonuses.

Additionally, the PR builds on @hillerstorm 's port of the simc "free roll-over" model, which adds extra DPS when refreshes occur immediately before a DoT tick, and implements the negative "munching" component of the same aura update delay mechanic. The munching part has no effect on Ret results as expected (the tests results change is entirely due to changing the delay distribution parameters for the positive effect), but reduces the value of the Rogue bonus quite significantly.

UI notices have also been added for all three 2-piece bonuses, to clarify that the munching / roll-over model requires PTR validation. If it turns out that Blizzard releases "munch-free" versions of these sets in Classic without the aura delays in place, then changing the sim implementation will be as simple as flipping one boolean flag in the Ignite configs, which was another motivation for cleaning up the shared code.


NerdEgghead:
@TheBackstabi Also tagging you for review

hillerstorm (approved):
🔥

PR #1046, opened 2024-09-22
Fix VPLC N damage range & class masks

1337LutZ:
- Adjust VPLC Normal damage according to database as tooltip is faulty
- Removed Balance of Power's incorrect class mask
- Removed Moonkin Form's incorrect class mask
- Add Balance Test

### DIFF
--- a/sim/common/cata/damage_procs.go
+++ b/sim/common/cata/damage_procs.go
@@ -66,7 +66,7 @@ func init() {
 			ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
 				// Tooltip is wrong:
 				// https://wago.tools/db2/SpellEffect?build=4.4.1.56574&filter[SpellID]=96887%7C96891%7C97119&page=1&sort[SpellID]=asc
-				baseDamage := sim.Roll(2560, 3292) * float64(dummyAura.GetStacks())
+				baseDamage := sim.Roll(2561, 3292) * float64(dummyAura.GetStacks())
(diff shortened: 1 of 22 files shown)


sim/common/cata/damage_procs.go
```diff
@@ -64,7 +64,9 @@ func init() {
 			ThreatMultiplier: 1,
 
 			ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
-				baseDamage := sim.Roll(985, 1266) * float64(dummyAura.GetStacks())
+				// Tooltip is wrong:
+				// https://wago.tools/db2/SpellEffect?build=4.4.1.56574&filter[SpellID]=96887%7C96891%7C97119&page=1&sort[SpellID]=asc
+				baseDamage := sim.Roll(2560, 3292) * float64(dummyAura.GetStacks())
```
raethkcj, line 69: When EffectDieSides is >= 1, the minimum roll is EffectBasePoints + 1, so 2561 here

sim/common/cata/damage_procs.go
```diff
@@ -111,7 +113,7 @@ func init() {
 			ThreatMultiplier: 1,
 
 			ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
-				baseDamage := sim.Roll(2889, 3713) * float64(dummyAura.GetStacks())
+				baseDamage := sim.Roll(2888, 3713) * float64(dummyAura.GetStacks())
```
raethkcj, line 116: And 2889 was correct here

PR #1049, opened 2024-09-22
[Item] Fix Stay of Execution to actually reduce dmg taken

hillerstorm:


NerdEgghead (approved):
Thanks for fixing this! There are a few things we'll still have to verify on PTR, such as whether the buff always lasts the full 30 seconds vs. the DoT starting to tick as soon as the absorb gets capped, but this is good enough as a first implementation.

PR #1050, opened 2024-09-22
[DK] Fix AMS runic power cost and dmg absorption amount

### LINKED ISSUE #887: [DEATH KNIGHT] AMS costs 20 runic power
**The bug**
Death knight spell Anti-Magic Shell costs 20 runic power instead of zero

**To Reproduce**
1. Select Unholy Death knight
2. Add Anti-Magic shell cast to the APL
3. Go to results > timeline and find the cast of AMS
4.  Mouseover the timestamp for runic power at the top and notice the cost from casting AMS

**Expected behavior**
It should cost nothing as is always the case in Cataclysm

_____
 - Simulations ran on browser version, Firefox Version 128.0.2 (64-bit)

hillerstorm:
Fixes #887 

Still needs a proper input and fake damage taken for UH to use as dps but this will at least fix the cost :)

### DIFF
--- a/sim/death_knight/anti_magic_shell.go
+++ b/sim/death_knight/anti_magic_shell.go
@@ -11,6 +11,7 @@ import (
 func (dk *DeathKnight) registerAntiMagicShellSpell() {
 	actionID := core.ActionID{SpellID: 48707}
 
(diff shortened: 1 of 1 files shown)


hillerstorm:
Closing in favor of #1082

PR #1051, opened 2024-09-23
[Rogue] Update default EPs, Breakpoints, and Gear Sets for Combat and Assassination

TheBackstabi:
Some minor optimizations of Asn and Combat gearsets, and some specific changes to reforge logic to try to avoid blowing past Spell Hit Cap, as it's typically a DPS loss compared to sitting just under.

Combat EPs specifically are a massive headache as the interactions between Hit, Haste, and Mastery result in many, MANY solutions that are _essentially_ equal with each other. You can increase/decrease the EP value of each by .1, get completely different reforge solutions, and the DPS difference will be within 15 or so - practically equal as far as in-game results are concerned.

My original intent was to have two distinct EP settings each prioritizing different things, but eventually settled on just one that has equivalent Haste/Mastery weights, and brings down the Spell Hit EP as it approaches some manually tested values (between 14 and 15%). This can likely be refined even further but it's really a situation that would require validating the reforge result with DPS instead of total EP value.


ui/rogue/combat/sim.ts
```diff
@@ -187,12 +188,41 @@ const SPEC_CONFIG = registerSpecConfig(Spec.SpecCombatRogue, {
 	],
 });
 
+function getMasterySoftCapConfig(): StatCap {
```
1337LutZ, line 191: Please make this an arrow function per modern ES6 standards;
`const getMasterySoftCapConfig = () => : StatCap`

ui/rogue/combat/sim.ts
```diff
@@ -187,12 +188,41 @@ const SPEC_CONFIG = registerSpecConfig(Spec.SpecCombatRogue, {
...
+	}
+
+	return StatCap.fromStat(Stat.StatMasteryRating, {
+		breakpoints: masteryRatingBreakpoints,
+		capType: StatCapType.TypeThreshold,
+		postCapEPs: [0],
+	});
+}
+
+function AddOrRemoveMasteryBreakpoint(optimizer: ReforgeOptimizer, isShown: boolean): void {
```
1337LutZ, line 205: Same here, and don't capitalise the function `addOrRemoveMasteryBreakpoint`

ui/rogue/combat/sim.ts
```diff
@@ -187,12 +188,41 @@ const SPEC_CONFIG = registerSpecConfig(Spec.SpecCombatRogue, {
...
+}
+
 export class CombatRogueSimUI extends IndividualSimUI<Spec.SpecCombatRogue> {
 	constructor(parentElem: HTMLElement, player: Player<Spec.SpecCombatRogue>) {
 		super(parentElem, player, SPEC_CONFIG);
 
 		player.sim.waitForInit().then(() => {
-			new ReforgeOptimizer(this);
+			const optimizer = new ReforgeOptimizer(this);
+			AddOrRemoveMasteryBreakpoint(optimizer, this.sim.getShowExperimental())
```
1337LutZ, line 221: Instead of binding a changeEmitter here I would just use the optimizer's `updateSoftCaps` callback function to enrich the Mastery breakpoints. You can check the `mage/fire/sim.ts` for an example. We modify it there, but you can also straight up just add an entry

1337LutZ (changes requested):

ui/rogue/combat/sim.ts
1337LutZ: **If** you wanted multiple EPs, you could also set that up. Example in the `warrior/fury/sim.ts`

TheBackstabi: Yeah I've gone back and fourth on having multiple EPs. After trying what I had on both T11 and T12 setups, I've settled on making one specific to having the complete 4PT12 as it's impactful enough to warrant being distinct.

1337LutZ: @TheBackstabi Ok, was just making sure you didn't do it because you thought it might not be possible in the current setup :D

1337LutZ (approved):

PR #1059, opened 2024-09-26
[Item] Add all phase 4 trinkets including item notices

hillerstorm:
Also changed Spidersilk Spindle to reduce incoming damage (like Savage Defense) instead of healing to be able to see its effects on incoming damage.
All proc masks and coefficients should be correct to how it is in the dbc and reading up on patch changes.

Added item notices for all of the procc trinkets, including the healing ones not being implemented.
Also added "not ingame" notices for the 384/410 variants of the Valor Trinkets which, just like some Firelands stuff, were never actually in game.


sim/common/cata/other_effects.go
```diff
@@ -825,6 +836,253 @@ func init() {
...
+				ProcChance: 0.45,
+				ICD:        time.Second * 9,
+				Handler: func(sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
+					lightningStrike.Cast(sim, result.Target)
+				},
+			})
+		})
+
+		fetishItemID := []int32{77982, 77210, 78002}[version]
+		core.NewItemEffect(fetishItemID, func(agent core.Agent) {
```
1337LutZ, line 888: Maybe add a note to the DBC value not existing and the patch changing it?

hillerstorm: Are you referring to Vial of Shadows? Yea I guess we could add a comment pointing to the patch notes, although we haven't tracked patch changes for other items afaik :D

1337LutZ: Ah yes sorry :D Since just like with VPLC we aren't 100% sure about the state this is implemented in before PTR

zoristaken:
About adding notices to items that were never in the game, I  think It should be more  correct, the moment we have confirmation that those versions won't exist in classic, to blacklist them instead of having them available as  a choice. Seems fine for now since we don't know yet if they will reuse them.

hillerstorm:
> About adding notices to items that were never in the game, I  think It should be more  correct, the moment we have confirmation that those versions won't exist in classic, to blacklist them instead of having them available as  a choice. Seems fine for now since we don't know yet if they will reuse them.

Yea that's my thinking as well. As soon as we get confirmation of blizzard not adding the upgraded firelands items we should blacklist them etc.
The text could perhaps be even more detailed, explaining a bit more.

1337LutZ:
> > About adding notices to items that were never in the game, I  think It should be more  correct, the moment we have confirmation that those versions won't exist in classic, to blacklist them instead of having them available as  a choice. Seems fine for now since we don't know yet if they will reuse them.
> 
> Yea that's my thinking as well. As soon as we get confirmation of blizzard not adding the upgraded firelands items we should blacklist them etc. The text could perhaps be even more detailed, explaining a bit more.

Yea I agree.

sim/common/cata/other_effects.go
```diff
@@ -8,6 +8,14 @@ import (
 	"github.com/wowsims/cata/sim/core/stats"
 )
 
+type ItemVersion int32
+
+const (
+	ItemVersionLFR    ItemVersion = 0
+	ItemVersionNormal ItemVersion = 1 << iota
+	ItemVersionHeroic
+)
```
NerdEgghead, line 17: Minor optimization, but you can do this instead:
```
type ItemVersion byte

const (
	ItemVersionLFR ItemVersion = iota
	ItemVersionNormal
	ItemVersionHeroic
)
```

This also lets you directly index arrays by it below.

sim/common/cata/other_effects.go
```diff
@@ -762,6 +755,24 @@ func init() {
...
+
+					if sim.Log != nil {
+						character.Log(sim, "Loom of Fate absorbed %.1f damage", absorbedDamage)
+					}
+
+					if currentShield <= 0 {
+						shield.Deactivate(sim)
+					}
+				}
+			})
```
NerdEgghead, line 774: Since we're now using this logic multiple times for trinkets, as well as for Savage Defense, you could consider adding a helper utility in core for it. For example,

```
func (character *Character) NewDamageAbsorptionAura(auraLabel string, actionID ActionID, duration time.Duration, shieldStrengthPtr *float64) *Aura {}
```

sim/common/cata/other_effects.go
```diff
@@ -825,6 +836,253 @@ func init() {
 			})
 		})
 	}
+
+	for version, _ := range []ItemVersion{ItemVersionLFR, ItemVersionNormal, ItemVersionHeroic} {
```
NerdEgghead, line 840: If you make the `ItemVersion` change, then this would be instead be
```
for _, version, := range []ItemVersion{ItemVersionLFR, ItemVersionNormal, ItemVersionHeroic} {
```

sim/common/cata/stat_bonus_stacking.go
```diff
@@ -174,4 +174,60 @@ func init() {
 		Harmful:    false,
 		ProcChance: 1,
 	})
+
+	for version, _ := range []ItemVersion{ItemVersionLFR, ItemVersionNormal, ItemVersionHeroic} {
```
NerdEgghead, line 178: same here

sim/common/cata/other_effects.go
NerdEgghead: Thinking about this more, we can even avoid the awkwardness of passing a pointer to the shield strength value if we embed it in a struct. For example:

```
type ShieldStrengthCalculator func(unit *Unit) float64

type DamageAbsorptionAura struct {
    *Aura
    ShieldStrength float64
    FreshShieldStrengthCalculator ShieldStrengthCalculator
}

func (aura *DamageAbsorptionAura) Activate(sim *Simulation) {
    aura.Aura.Activate(sim)
    aura.ShieldStrength = aura.FreshShieldStrengthCalculator(aura.Unit)
}
```

Then we could have the helper be constructed as

```
func (character *Character) NewDamageAbsorptionAura(auraLabel string, actionID ActionID, duration time.Duration, calculator ShieldStrengthCalculator) *DamageAbsorptionAura
```

The downstream code that calls this would then hold onto that returned pointer, and would simply have to call `Activate()` on it as part of either the trinket activation spell or the `ProcTrigger` handler, depending on how the shield is triggered.

hillerstorm:
Updated the item-doesnt-exist-notice a bit:
image

sim/common/cata/other_effects.go
```diff
@@ -825,6 +786,207 @@ func init() {
...
+	for _, version := range []ItemVersion{ItemVersionLFR, ItemVersionNormal, ItemVersionHeroic} {
+		labelSuffix := []string{" (LFR)", "", " (Heroic)"}[version]
+
+		vialItemID := []int32{77979, 77207, 77999}[version]
+		core.NewItemEffect(vialItemID, func(agent core.Agent) {
+			character := agent.GetCharacter()
+
+			actionID := core.ActionID{SpellID: []int32{109721, 107994, 109724}[version]}
+			minDmg := []float64{3568, 4028, 4546}[version]
+			maxDmg := []float64{5353, 6042, 6819}[version]
```
hillerstorm, line 799: So what do we do with [Vial of Shadows](https://www.wowhead.com/cata/item=77999/vial-of-shadows)? Simc has a 0.226, 0.3, 0.339 ap scaling here while the dbc is missing it atm.

Leaving the ap scaling out turns this trinket extremely lackluster and adding it is >1k more dps 😅 with ap scaling it does about as much dmg over 180s for a feral as Fetish does for an Arms warr which seems legit?

This comment from simc worries me though that it wasn't in the dbc even back then 🤷:
> // ICD, base damage, and ap coeff determined experimentally on heroic version. Assuming dbc has wrong base damage. Normal and LFR assumed to be changed by the same %

Thoughts @NerdEgghead @zoristaken @RobFalfa @1337LutZ ?

1337LutZ: I'd go for the underwhelming implementation with the increased proc rate according to the "nerf"/adjustment it got and add an item notice

hillerstorm:
I added a TODO comment on Vial of Shadows regarding the AP modifier with some links for reference later.

Would you feel comfortable with merging this @NerdEgghead seeing as I touched Savage Defense? :) that change didn't result in any test diffs and from my own testing it looked great.

PR #1062, opened 2024-09-27
[UI][WARLOCK] Update Warlock P3 setups

1337LutZ:
- Added preliminary P3 setups for Warlock specs
- Added Demonology Incinerate spec


Tharre (changes requested):
A couple things:

- Don't test every gear set, just the newest one. Tests are already slower then they should be, and testing every preset adds significantly to that problem
- Same with destro P1 & P3 rotations, probably want to keep the incinerate one though since it's utilizing a different spell entirely
- Actually the entire destro P1 rotation preset is a bit suspect to me, the P3 one sims the same even for P1 gear (with fel flame)
- I don't think adding Glyph of Felhunter as default is a good idea; it's finicky getting it to work, not like normal glyphs that are set & forget. Users should opt into it if they want to use it
- Can you format the json for the APLs like the were before, i.e. one action per line instead of "pretty" printing them, makes it much harder to read & edit

1337LutZ:
> * Don't test every gear set, just the newest one. Tests are already slower then they should be, and testing every preset adds significantly to that problem
> * Same with destro P1 & P3 rotations, probably want to keep the incinerate one though since it's utilizing a different spell entirely
> * Actually the entire destro P1 rotation preset is a bit suspect to me, the P3 one sims the same even for P1 gear (with fel flame)
> * I don't think adding Glyph of Felhunter as default is a good idea; it's finicky getting it to work, not like normal glyphs that are set & forget. Users should opt into it if they want to use it
> * Can you format the json for the APLs like the were before, i.e. one action per line instead of "pretty" printing them, makes it much harder to read & edit

- **Don't test every gear set, just the newest one. Tests are already slower then they should be, and testing every preset adds significantly to that problem**
  - Done
- **Same with destro P1 & P3 rotations, probably want to keep the incinerate one though since it's utilizing a different spell entirely**
  - Done and Left in the Demo rotations/talents
- **Actually the entire destro P1 rotation preset is a bit suspect to me, the P3 one sims the same even for P1 gear (with fel flame)**
  - You are right. Removed the P1 and left in the P3 Chaos bolt changes + Added the Fel Flame line. Renamed to Default.
- **I don't think adding Glyph of Felhunter as default is a good idea; it's finicky getting it to work, not like normal glyphs that are set & forget. Users should opt into it if they want to use it**
  - Ok, reverted it. Just saw the backend tests doing it, so thought it was maybe an oversight, esp since the Discord always asks for FH.
- **Can you format the json for the APLs like the were before, i.e. one action per line instead of "pretty" printing them, makes it much harder to read & edit**
  - Done

Tharre (approved):
You missed the affliction APL, otherwise LGTM

1337LutZ:
> You missed the affliction APL, otherwise LGTM

Sorry mb, I have a habit of pressing auto format...

PR #1063, opened 2024-09-27
[MAGE] T13 set bonus

1337LutZ:


sim/mage/combustion.go
```diff
@@ -36,6 +36,9 @@ func (mage *Mage) registerCombustionSpell() {
 				spell.DealDamage(sim, result)
 				spell.RelatedDotSpell.Cast(sim, target)
 			}
+			if mage.t13ProcAura != nil {
+				spell.CD.Reduce(time.Second * time.Duration(5*mage.t13ProcAura.GetStacks()))
+			}
```
NerdEgghead, line 41: This should go in the `OnStacksChange` callback for `mage.t13ProcAura` rather than in the `ApplyEffects` for the affected spell. With your current implementation, if more stacks are gained after Combustion is cast, the CD won't be reduced further when it should.

sim/mage/combustion.go
```diff
@@ -82,6 +85,9 @@ func (mage *Mage) registerCombustionSpell() {
 			},
 			OnTick: func(sim *core.Simulation, target *core.Unit, dot *core.Dot) {
 				dot.CalcAndDealPeriodicSnapshotDamage(sim, target, dot.OutcomeSnapshotCrit)
+				if mage.t13ProcAura != nil && !dot.IsActive() {
+					mage.t13ProcAura.SetStacks(sim, 0)
+				}
```
NerdEgghead, line 90: It would be cleaner to put this in the `OnExpire` of the Aura config for the DoT.

sim/mage/items.go
```diff
@@ -83,11 +84,47 @@ var ItemSetTimeLordsRegalia = core.NewItemSet(core.ItemSet{
...
+					Label:     "Stolen Time",
+					ActionID:  core.ActionID{SpellID: 105785},
+					Duration:  time.Second * 30,
+					MaxStacks: 10,
+				},
+				BonusPerStack: stats.Stats{stats.HasteRating: 50},
+			})
+
+			newStolenTimeTrigger := func(procChance float64, spellMask int64) {
+				core.MakePermanent(core.MakeProcTriggerAura(&character.Unit, core.ProcTrigger{
```
NerdEgghead, line 105: You don't need the `MakePermanent()` wrapper here, that's done by default if no `Duration:` is supplied in the `ProcTrigger` config.

sim/mage/items.go
```diff
@@ -83,11 +84,47 @@ var ItemSetTimeLordsRegalia = core.NewItemSet(core.ItemSet{
...
+					Name:           "Stolen Time Trigger",
+					ActionID:       core.ActionID{ItemID: 105788},
+					Callback:       core.CallbackOnCastComplete,
```
NerdEgghead, line 108: This might need to be `core.CallbackOnSpellHitDealt` to work properly?

sim/mage/talents_arcane.go
```diff
@@ -311,6 +311,9 @@ func (mage *Mage) registerArcanePowerCD() {
 			}
 			mage.arcanePowerCostMod.Deactivate()
 			arcanePowerDmgMod.Deactivate()
+			if mage.t13ProcAura != nil {
+				mage.t13ProcAura.SetStacks(sim, 0)
```
NerdEgghead, line 315: You should call `Deactivate()` here (which will automatically set the stacks to 0) rather than only `SetStacks()`.

NerdEgghead (changes requested):

sim/mage/combustion.go
1337LutZ: From what I've read it calculates it on casting the sprll, not after/during the uptime. So any stacks after casting this do not count anymore.

sim/mage/talents_arcane.go
1337LutZ: Would that not also remove the buff from the timeline until your next cast? The way I understood it, it will reset the buff to 0 stacks.
I guess it doesnt really matter for it to work and .Deactivate looks cleaner

NerdEgghead (approved):

PR #1065, opened 2024-09-30
[WARLOCK] Reverse UA glyph cast time logic

1337LutZ:
- The glyph was increasing the cast time instead of decreasing


Tharre (approved):
LGTM

PR #1089, opened 2024-10-08
Add custom APL Mage Combust value

1337LutZ:
- Added APL value for calculating the total Combustion Dot value
  - Simplifies the APL logic for Combustion
  - Mimics how Mage plays around Combustion in-game using their Combustion WeakAura
- When re-simming the active Log Search input value will be applied by default to filter the new results
- Add Troll and Non-Troll Simple Rotation presets based on plotted Combust Dot values (see graphs below)

**Simple rotation**
image

**APL**
image

**Plotted Combust values**
combined-avg
combust-troll
combust-worgen


1337LutZ:
Awaiting confirmation & testing from Mage Discord

InDebt (approved):

PR #1110, opened 2024-10-18
Add ability to limit breakpoint capping

1337LutZ:
- Add the ability to let the user choose to override to a max breakpoint limit
  - Example: A Mage would like to have more control over which breakpoint should be the cap (15%, 20% etc), to go a bit more Crit heavy. The stat cap flow doesn't work for that due to 0-ing the EP after reaching a cap.
- Dropdown is based on the available options defined by the Sim

image

### DIFF
--- a/ui/core/components/suggest_reforges_action.tsx
+++ b/ui/core/components/suggest_reforges_action.tsx
@@ -57,6 +57,8 @@ export type ReforgeOptimizerOptions = {
 	experimental?: true;
 	statTooltips?: StatTooltipContent;
 	statSelectionPresets?: UnitStatPresets[];
+	// Allows you to enable breakpoint limits for Treshold type caps
+	enableBreakpointLimits?: boolean;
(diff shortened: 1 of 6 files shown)


ui/core/components/suggest_reforges_action.tsx
```diff
@@ -625,14 +628,105 @@ export class ReforgeOptimizer {
 		);
 	}
 
+	buildSoftCapBreakpointsLimiter({ useSoftCapBreakpointsInput }: { useSoftCapBreakpointsInput: BooleanPicker<Player<any>> | null }) {
+		if (!useSoftCapBreakpointsInput) return null;
```
NerdEgghead, line 632: Does this also need a check for `this.enableBreakpointLimits`?

1337LutZ: Yep, you're right. Added!

PR #1111, opened 2024-10-19
[Feature] Apl Aura/CD Sets

NerdEgghead:
This PR aims to dramatically improve the readability of snapshotting APLs by creating wrappers for (a) casting all major cooldowns that buff a desired set of stats, and (b) performing common checks on any equipped proc trinkets that buff a desired set of stats. These wrappers eliminate the need to exhaustively list every single on-use or proc trinket trinket in the game when constructing a gear-agnostic APL, and make the core rotational decisions for snapshotting much clearer to the reader as a result.


raethkcj:
Thanks a lot for looking at this! Personally I feel the most flexible and helpful goal for something like this would be, instead of providing a top-level APL action, provide a value in `castable_spells` akin to [`OtherActionPotion`](<https://github.com/wowsims/cata/blob/72dcc20f317a2842502b0b249bad9262c3fc696f/ui/core/components/individual_sim_ui/apl_helpers.tsx#L109>), except instead of [breaking](https://github.com/wowsims/cata/blob/72dcc20f317a2842502b0b249bad9262c3fc696f/sim/core/apl_helpers.go#L139) on the first match, do what [`processMajorCooldowns`](https://github.com/wowsims/cata/pull/1111/files#diff-545fbaa607e343669f9ef5b38e7d95e982cbfcfe81d6942da9101653b986d3a0R291-R294) is doing with generating subactions for each spell from `GetMatchingStatBuffSpells`. But instead of hardcoding `APLActionCastSpell`, use whatever the actual action is.

That's my vision, anyway. I know it's a lot, especially when it comes to auras as values in conditions.
.

NerdEgghead:
> Thanks a lot for looking at this! Personally I feel the most flexible and helpful goal for something like this would be, instead of providing a top-level APL action, provide a value in `castable_spells` akin to [`OtherActionPotion`](https://github.com/wowsims/cata/blob/72dcc20f317a2842502b0b249bad9262c3fc696f/ui/core/components/individual_sim_ui/apl_helpers.tsx#L109), except instead of [breaking](https://github.com/wowsims/cata/blob/72dcc20f317a2842502b0b249bad9262c3fc696f/sim/core/apl_helpers.go#L139) on the first match, do what [`processMajorCooldowns`](https://github.com/wowsims/cata/pull/1111/files#diff-545fbaa607e343669f9ef5b38e7d95e982cbfcfe81d6942da9101653b986d3a0R291-R294) is doing with generating subactions for each spell from `GetMatchingStatBuffSpells`. But instead of hardcoding `APLActionCastSpell`, use whatever the actual action is.
> 
> That's my vision, anyway. I know it's a lot, especially when it comes to auras as values in conditions.
> 
> EDIT: in the case of auras, traverse the tree upward to nearest Any/All Of value and generate everything below that for each aura?

Making this work for any arbitrary query you could possibly make on a set of auras is far too onerous a task to undertake, but feel free to take a stab at if you want ;)

I think what I've put together should already cover the common snapshotting cases, and it's easy enough to add new values as the need arises.

sim/core/apl_actions_casting.go
```diff
@@ -313,7 +313,18 @@ func (action *APLActionCastAllStatBuffCooldowns) Execute(sim *Simulation) {
...
+}
+func (action *APLActionCastAllStatBuffCooldowns) PostFinalize(rot *APLRotation) {
+	if len(action.allSubactions) == 0 {
+		rot.ValidationWarning("%s will not cast any spells! There are either no major cooldowns buffing the specified stat type(s), or all of them are manually cast in the APL.", action)
+	} else {
+		actionIDs := MapSlice(action.allSubactions, func(subaction *APLActionCastSpell) ActionID {
+			return subaction.spell.ActionID
+		})
+
+		rot.ValidationWarning("%s will cast the following spells: %s", action, StringFromActionIDs(actionIDs))
```
1337LutZ, line 328: Could we not leave in the action IDs and then in frontend map them to actual spells with icons? Like we do in the logs parser? Could maybe be cleaner/easier to visually reference for users?
Or perhaps list them on a new line?

NerdEgghead: It should already be printing the exact same string as it does for single spells, just separated by commas. Is there something I need to hook into in the UI code to pretty-print it? It's already recognizing the ActionID string automatically and cleanly displaying just the spell name, just without the icon etc.

1337LutZ: Yea, in the UI we would probably need to add a regex that parses the validation message > makes an ActionID out of its > replaces it with a `<a href>` and initialize the wowhead tooltips code. But that can also be a future state if needed. 
Actually, I found it already being used in the APL here: https://github.com/wowsims/cata/blob/4913dd7a9e209b427ee1acab40e8bd45eb6eb8f3/ui/core/components/individual_sim_ui/apl_rotation_picker.ts#L266

NerdEgghead: Yeah it's clearly already doing some parsing automatically, or else it wouldn't be printing the spell names cleanly (the direct %s of an ActionID struct does not look like that)

1337LutZ: Ohhhh, my bad then 👍 feel free to merge regardless

NerdEgghead:
I think I'll PR the Unholy migration (along with any new APL values required to support it) separately, since they're still working on getting me a sufficiently generalized brute force example to work from. So will just merge the two caster spec migrations once the spec owners sign off on it.

Polynomix (approved):
I found an UI issue, going from the base default rotation and disabling the Time To Ready FET > 90s rule mess with the new warnings. It moves it up and down ?
image
image
I don't know if its a general warning thing or if its linked to this
Otherwise the new snapshottting rule seems to behave the same. The way the volcanic potion is used for 2nd Fire ele is still weird but that's just because of the inconsistency in the 2pt12 procs.

NerdEgghead:
> I found an UI issue, going from the base default rotation and disabling the Time To Ready FET > 90s rule mess with the new warnings. It moves it up and down ? image image I don't know if its a general warning thing or if its linked to this Otherwise the new snapshottting rule seems to behave the same. The way the volcanic potion is used for 2nd Fire ele is still weird but that's just because of the inconsistency in the 2pt12 procs.

That bug was already present beforehand, but the new warnings will also play into the same bug.  I know where it's happening in core but ignored it because it'll be an annoying fix on the back-end. There might be a simpler fix for it on the front-end though, I can take a look.

sim/core/apl_helpers.go
```diff
@@ -127,6 +128,18 @@ func (rot *APLRotation) GetAPLICDAura(sourceUnit UnitReference, auraId *proto.Ac
...
 }
 
+func (rot *APLRotation) GetAPLTrinketProcAuras(statTypesToMatch []stats.Stat, warnIfNoneFound bool) []*StatBuffAura {
+	unit := rot.unit
+	character := unit.Env.Raid.GetPlayerFromUnit(unit).GetCharacter()
+	matchingAuras := character.GetMatchingTrinketProcAuras(statTypesToMatch)
+
+	if (len(matchingAuras) == 0) && warnIfNoneFound {
+		rot.ValidationWarning("No trinket proc buffs found for: %s", StringFromStatTypes(statTypesToMatch))
+	}
```
raethkcj, line 138: Is it feasible to "warn" the list of auras that are found, like the cast ones are doing?

sim/shaman/elemental/TestElemental.results
```diff
@@ -1012,15 +1012,15 @@ dps_results: {
...
+  dps: 34203.82708
+  tps: 596.51235
  }
 }
 dps_results: {
  key: "TestElemental-AllItems-MandalaofStirringPatterns-62472"
  value: {
-  dps: 34363.35035
-  tps: 597.13505
+  dps: 34203.82708
```
raethkcj, line 1022: Interesting that a couple items went down, will try to double check what happened. The ones that went up should mostly be low-ilvl or healer-stat ones that I excluded from the original list to keep it a bit shorter.

raethkcj (commented):
Selecting Strength in any `statTypeFieldConfig` sets it to None; all the other stats work, maybe some kind of index error?

sim/shaman/elemental/TestElemental.results
raethkcj: Oh right it's just RNG having different outcomes since the cast order changed, disregard

NerdEgghead:
> Selecting Strength in any `statTypeFieldConfig` sets it to None; all the other stats work, maybe some kind of index error?

This should be fixed now, good catch!

sim/core/apl_helpers.go
NerdEgghead: This should be implemented now.

raethkcj (approved):
Thanks a ton, looks good to me! Hopeful that the validation *cough* abuse *cough* can be revisited sometime down the road, but no need for that to fall on you, and it's a big leap forwards either way.

PR #1125, opened 2024-10-24
Add Gurthalak, Voice of the Deeps

hillerstorm:
Tried to replicate all the EJ-, Wowhead- and Simc-comments I could gather.
Damage matches all the videos of the different versions of the sword on YouTube, including which buffs increase the proc damage.

Had to modify Unholy mastery to apply to all shadow damage and move the SpellMod to only modify Unholy Blight and Scourge Strike off-hand since they had IgnoreAllModifiers.

Also made a change to Death Wish for Fury Warriors since it's only supposed to increase physical damage, changes most tests a big seeing as it appeared to modify the T12 fire-proc.

I've been restrictive with the proc mask for now and skipped the `Can Proc From Procs` flag, it's there in the db but it already pretty much matches the expected numbers.
The rest has to be tested on the PTR :)
Left lot's of TODOs for future testing!

The cap for amount of simultaneous spawns is 10. I tested with 500k 15min Fury TG sims with added haste and white hit cap and still couldn't get more than 9 to spawn so it should be safe.

### DIFF
--- a/sim/common/cata/gurthalak.go
+++ b/sim/common/cata/gurthalak.go
@@ -1,0 +1,238 @@
+package cata
+
+import (
+	"time"
+
+	"github.com/wowsims/cata/sim/core"
+	"github.com/wowsims/cata/sim/core/stats"
+)
+
+// Gurthalak, Voice of the Deeps
+// Equip: Your melee attacks have a chance to cause you to summon a Tentacle of the Old Ones to fight by your side for 12 sec.
+// (Proc chance: 2%)
+func init() {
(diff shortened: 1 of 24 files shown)


InDebt:
lgtm - a shame though we can't bind the pets to the actual Item Effect but have to init them in the Agent constructor.

hillerstorm:
> lgtm - a shame though we can't bind the pets to the actual Item Effect but have to init them in the Agent constructor.

Yea, would've been much cleaner but it is what it is 😅

PR #1135, opened 2024-10-27
[Balance] Adjustments to T12 2P based on PTR data

stn-ms:


sim/druid/burning_treant.go
```diff
@@ -16,10 +16,16 @@ type BurningTreant struct {
 }
 
 func (druid *Druid) NewBurningTreant() *BurningTreant {
-	var burningTreantBaseStats = stats.Stats{stats.SpellCritPercent: 5}
+	baseStats := stats.Stats{stats.SpellCritPercent: 0}
```
1337LutZ, line 19: Ah yea, I guess it's the same findings I had with Fiery Imp from Warlock 2P

stn-ms: Out of all the data we gathered, there isn't a single log where fireseed has crit.

1337LutZ: Was this tested without the 5% raid debuff?

1337LutZ: Was just double checking. both Warlock and Mage show 0 base crit and 0 inherited crit 👍

PR #1136, opened 2024-10-27
[Elemental] Some changes to dtr aoe spells and bug fixes

Polynomix:
- Dtr procs do not consume clearcasting and proc feedback
- Changed Bell of Enraging Resonance to proc on procs
- Revert DMC proccing on procs
- Earthquake damage event is a proc
- Adjusted the interaction of some AoE spells with DTR

Core DTR : Switched to comparing spells by their reference. If two spells had the same spell id, they couldn't both get duplicated
                 When using the ProcPerCast thing i have multithreading issue, adding an OnReset fixes it.


sim/shaman/elemental/dragonwrath.go
```diff
@@ -39,11 +47,29 @@ func customFulminationHandler(sim *core.Simulation, spell *core.Spell, result *c
...
+	baseDamage := shaman.ClassSpellScaling*0.32400000095 + 0.11*spell.SpellPower()
+
+	if copySpell == nil {
+		copyConfig := cata.GetDRTSpellConfig(spell)
+		copyConfig.Cast = core.CastConfig{}
+		copyConfig.Flags |= core.SpellFlagIgnoreResists // Doesnt get double damage reduced by armor
+		copySpell = spell.Unit.RegisterSpell(copyConfig)
+	}
+
+	copySpell.ApplyEffects = damageFactory(baseDamage)
```
InDebt, line 67: This would cause EQ to only replicate damage to the target instead of all targets. See above, but I think the whole 'proc' is replicated? Or am I miss reading the log and EQ overlapped for some reason? Did not find many logs tbh

sim/shaman/elemental/dragonwrath.go
```diff
@@ -39,11 +47,29 @@ func customFulminationHandler(sim *core.Simulation, spell *core.Spell, result *c
...
+func customEarthquakeHandler(sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
+	copySpell := spell.Unit.GetSpell(spell.WithTag(71086))
+
+	shaman := spell.Unit.Env.Raid.GetPlayerFromUnit(spell.Unit).(shaman.ShamanAgent).GetShaman()
+	baseDamage := shaman.ClassSpellScaling*0.32400000095 + 0.11*spell.SpellPower()
+
+	if copySpell == nil {
+		copyConfig := cata.GetDRTSpellConfig(spell)
+		copyConfig.Cast = core.CastConfig{}
+		copyConfig.Flags |= core.SpellFlagIgnoreResists // Doesnt get double damage reduced by armor
```
InDebt, line 63: I might be missing something, but why is the damage reduced by armor twice? The ApplyEffects function calculates the damage again based on the base damage given. Which should be equivalent to the inital cast's way of doing it, as your're not basing of the new damage on the old one, there should be no second armor reduction, it would not simply ignore armor, no?

sim/shaman/elemental/dragonwrath.go
```diff
@@ -8,12 +8,19 @@ import (
...
+		AddSpell(88767, cata.NewDragonwrathSpellConfig().WithSpellHandler(customFulminationHandler)).                              // Fullmination
+		AddSpell(403, cata.NewDragonwrathSpellConfig().WithCustomSpell(overloadCopyHandler)).                                      // Lightning Bold
+		AddSpell(421, cata.NewDragonwrathSpellConfig().WithCustomSpell(overloadCopyHandler).ProcPerCast()).                        // Chain Lightning
+		AddSpell(51505, cata.NewDragonwrathSpellConfig().WithCustomSpell(overloadCopyHandler)).                                    // Lava Burst
+		AddSpell(3599, cata.NewDragonwrathSpellConfig().SupressSpell()).                                                           // Searing Totem
+		AddSpell(8190, cata.NewDragonwrathSpellConfig().SupressSpell()).                                                           // Magma Totem
+		AddSpell(51490, cata.NewDragonwrathSpellConfig().ProcPerCast()).                                                           // Thunderstorm
+		AddSpell(77478, cata.NewDragonwrathSpellConfig().IsAoESpell().TreatTickAsCast().WithSpellHandler(customEarthquakeHandler)) // Earthquake
```
InDebt, line 23: Looking at how EQ is implemented I'd suggest that you don't even need any fancy custom handling from DTR.
You could simply remodel the spell like I did for Druid's mushrooms here:

https://github.com/wowsims/cata/commit/fea603605283b8e1bcf9e9fdd3739ad7255a7841#diff-193a3c2777a162ceb4599f49352b2c5cbd6897dbe5589811348df4490871590c

As from my obersevation each tick is simply a an AoE cast like mind sear or shrooms that gets replicated as a whole. You'll still need to mark it as TreatTickasCast when you want to deal PeriodicDamage with it though

https://classic.warcraftlogs.com/reports/aHYJCbRBzK3rFfqT#boss=-3&difficulty=0&type=damage-done&start=7555099&end=7558099&ability=77478&view=events

Should show that the whole tick is replicated.

sim/shaman/elemental/dragonwrath.go
Polynomix: If you're talking about my screenshot in the ele dtr issue, i should have mentionned that it was on 3 targets. Each damage can duplicate independently of each other. For a same "pulse" of EQ, you can have 0 1 2 3 dup. And it only does damage to the target it procs on

sim/shaman/elemental/dragonwrath.go
Polynomix: It might not be reduced twice, i saw that the damage of the replicated spells was doing less damage and assumed that. Adding this is fixing it (in the debug logs, it was at the AfterTargetResistance that damage dropped), but you're right the issue has to be something else, i'm gonna look more into it

sim/shaman/elemental/dragonwrath.go
Polynomix: i think he got lucky and had 3 replicated tick at once. here's my log from testing : https://classic.warcraftlogs.com/reports/XckVzbHgYyD6JtKP#boss=-3&difficulty=0&source=22&type=damage-done&ability=77478&start=1848649&end=2529940&view=events on 3 targets

Polynomix: > Should show that the whole tick is replicated.

In fact, its not even the same targets hit by the replicated spells, i'm gonna ask him where he was testing

sim/shaman/elemental/dragonwrath.go
InDebt: Yeah I missed that one and looked out for an EQ log myself. Looking at your proc chance in your log at about 2.6%, thatwould somewhat align with the proc rate in the sim right now for AoE spells for shamans.

The 3x replication occurence would be 2.6%^3 however which is like 0.00176% or a one in 57k occurrence. So either lucky, or might there be a possiblity that the spell gets duped whole but from the effected target as the source? Not sure where you tested.

Polynomix: that might be a thing yes, i was doing this 
image
that would also explain why in the other log, there is a new target out of nowhere

sim/shaman/elemental/dragonwrath.go
Polynomix: Phys dot are ignoring armor, so as the base spell is implemtented as a dot and the replicated one not, the damage is not the same without this flag

InDebt (approved):

PR #1143, opened 2024-10-30
Standardize default fight length across sims

NerdEgghead:
On branch unify-fight-length
 Changes to be committed:
	modified:   ui/core/encounter.ts
	modified:   ui/core/individual_sim_ui.tsx
	modified:   ui/druid/feral/presets.ts
	modified:   ui/hunter/beast_mastery/presets.ts
	modified:   ui/hunter/marksmanship/presets.ts
	modified:   ui/hunter/survival/presets.ts
	modified:   ui/paladin/retribution/presets.ts
	modified:   ui/rogue/assassination/presets.ts
	modified:   ui/rogue/combat/presets.ts
	modified:   ui/rogue/subtlety/presets.ts
	modified:   ui/shaman/elemental/presets.ts
	modified:   ui/shaman/enhancement/presets.ts
	modified:   ui/warlock/affliction/presets.ts
	modified:   ui/warlock/demonology/presets.ts
	modified:   ui/warlock/destruction/presets.ts

### DIFF
--- a/ui/core/encounter.ts
+++ b/ui/core/encounter.ts
@@ -10,8 +10,8 @@ import { EventID, TypedEvent } from './typed_event';
 export class Encounter {
 	readonly sim: Sim;
 
-	private duration = 180;
-	private durationVariation = 5;
+	private duration = 300;
+	private durationVariation = 60;
 	private executeProportion20 = 0.2;
 	private executeProportion25 = 0.25;
 	private executeProportion35 = 0.35;
@@ -157,8 +157,8 @@ export class Encounter {
 		this.fromProto(
 			eventID,
 			EncounterProto.create({
-				duration: 180,
-				durationVariation: 5,
+				duration: 300,
+				durationVariation: 60,
 				executeProportion20: 0.2,
 				executeProportion25: 0.25,
 				executeProportion35: 0.35,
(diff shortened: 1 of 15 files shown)


NerdEgghead:
Relevant discussion threads on this topic:

https://discord.com/channels/891730968493305867/1215392965703508049/1300036065234653194
https://discord.com/channels/891730968493305867/1034670402150092841/1300815044241657859

hillerstorm (approved):

1337LutZ (approved):

Polynomix:
For ele sims, you don't glyph the same for 240s fight and 360s fight. But i'm fine with this since i believe people who cares enough about it will run their own duration with the setup they want.

ToxicKevinFerm (approved):

NerdEgghead:
> This would not affect any existing session any user has for Wowsims, right? As encounter length is to my knowledge not tied to the t11 or t12 prests, users would only get this by 'Restoring defaults', right?
> 
> Is there a way we can use something similar tot he stat changes to 'force' update the default fight settings for this version once?

There are some hacky ways to force this, but I don't think it's a good idea personally. Changing people's stored settings without a prompt isn't good practice unless it's truly required for the sim to run properly (stats indexing changes etc.). If someone had a custom fight length they were optimizing for but hadn't saved it as a preset, a forced reset to defaults will be very annoying for that user. Better practice is to just recommend that people restore to defaults on class Discords etc. That's just my two cents on this anyhow.

If we move towards my "Option 2" idea, then what we could do is utilize the Preset Builds functionality that @1337LutZ put in (which I heavily use for Guardian), and make sure that every spec always has a Preset Build with just the default encounter + fight length recommendation for the current phase saved in it. Then there's a button on that tab which a user can click to load all encounter defaults without needing to go into the global settings menu and restore everything.

PR #1162, opened 2024-11-03
[Rogue] Implement MH weapon optimization for Human/Orc Combat Rogues.

TheBackstabi:
Currently these two races will reforge for the full 6.5% Expertise, ignoring the MH weapon type. In-game, this results in 29/26 Expertise, wasting 90 rating that the MH doesn't benefit from.

This will make it reforge for 5.75% when the MH type matches their racial. In most cases, this is a ~90 DPS increase - fairly minor, but it's also free DPS.

### DIFF
--- a/ui/rogue/combat/sim.ts
+++ b/ui/rogue/combat/sim.ts
@@ -6,7 +6,7 @@ import { IndividualSimUI, registerSpecConfig } from '../../core/individual_sim_u
 import { Player } from '../../core/player';
 import { PlayerClasses } from '../../core/player_classes';
 import { APLRotation } from '../../core/proto/apl';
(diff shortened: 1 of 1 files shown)


1337LutZ:
Can we not just fix the fact that the Sim wont detect the bonus expertise? :O Afaik it works for Warr

TheBackstabi:
The problem is that Combat Rogues, specifically, will _never_ have a matching MH and OH weapon type (outside of Lego daggers). If both weapons are the same type, works fine. But because Combat uses a non-dagger MH and a 1.4 dagger OH, the current method of how we add in that bonus expertise doesn't work.

If we wanna fix the underlying detection and make the reforge logic examine just the MH, that would work too, but I also have no idea where that happens 😆

1337LutZ (approved):

PR #1166, opened 2024-11-04
Blacklist proc trinkets with extra proc conditions from APL aura sets

NerdEgghead:
Currently only affects the Sorrowsong trinket.

### DIFF
--- a/sim/common/cata/stat_bonus_procs.go
+++ b/sim/common/cata/stat_bonus_procs.go
@@ -321,7 +321,7 @@ func init() {
 		ICD:        time.Second * 50,
 	})
 
-	shared.NewProcStatBonusEffectWithCustomCondition(shared.ProcStatBonusEffect{
+	shared.NewProcStatBonusEffect(shared.ProcStatBonusEffect{
 		Name:       "Sorrowsong",
 		ID:         55879,
 		AuraID:     90990,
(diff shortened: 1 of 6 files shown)


Polynomix:
This will blacklist entirely sorrowsong. It can still proc under 35% hp and should be included in the tests then. If not, it doesn't solve the problem that the apl will include specific handling for sorrowsong. There is also 2 other tank trinkets that uses custom conditions, I'm not a tank expert but they might encounter the same issue if someone decides to equip one.
What do you think of testing if the extra condition is satisfied [here](https://github.com/wowsims/cata/blob/501e8d37803416a598c5c3f758e7c50386084600/sim/core/apl_values_aura_sets.go#L73) ?

NerdEgghead:
> This will blacklist entirely sorrowsong. It can still proc under 35% hp and should be included in the tests then. If not, it doesn't solve the problem that the apl will include specific handling for sorrowsong. There is also 2 other tank trinkets that uses custom conditions, I'm not a tank expert but they might encounter the same issue if someone decides to equip one.
> What do you think of testing if the extra condition is satisfied [here](https://github.com/wowsims/cata/blob/501e8d37803416a598c5c3f758e7c50386084600/sim/core/apl_values_aura_sets.go#L73) ?

Took some refactoring, but I think it should be behaving the way you want now.

Polynomix (approved):

PR #1172, opened 2024-11-05
[Ret] Combine and optimize APLs and add more pvp trinkets

### LINKED ISSUE #1170: Same Trinket, Higher Stats, 1000 lower dps?
[Vicious Gladiator's Insignia of Victory] level 371 gives 1000dps lower than same [Vicious Gladiator's Insignia of Victory] level 365 using no reforge?

hillerstorm:
* Expose Glyph of Exorcism and T13 2pc as auras to be able to use in APL
* Unify Default and T13 APLs into one using the above mentioned auras
* Tweak APLs for extremely minor gains
* Add all the remaining missing PvP trinkets (fixes #1170)


zoristaken:
I assume this fixes #1170 ?

hillerstorm:
> I assume this fixes #1170 ?

Oh, that's correct! Didn't look at the issue list 😄

PR #1173, opened 2024-11-06
Prepull APL action to emulate gear swapping cd

Polynomix:
People were asking for a way to trigger the 30s cooldown on spells that are associated with gear (on use trinkets, synapse) when that gear is equipped. This is used to emulate a prepull where different trinkets/gear is being used than the one used for the sim.

### DIFF
--- a/proto/api.proto
+++ b/proto/api.proto
@@ -431,6 +431,7 @@ message SpellStats {
 	bool encounter_only = 8; // Whether this spell may only be cast during the encounter (not prepull).
 	bool has_cast_time = 9; // Whether this spell has a cast time or not.
 	bool is_friendly = 10; // Whether this spell should be cast on player units
+	bool has_gscd = 11; // Wether this spell is associated to an item that incurs a 30s cd when equipped.
 }
 message APLActionStats {
 	repeated string warnings = 1;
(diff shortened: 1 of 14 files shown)


NerdEgghead:
Do we really need every `Spell` struct to hold a new CD timer for this to work? Wouldn't it be cleaner to just implement a generalized action like `SetSpellCD()`, which takes in both the spell and the time it should come off cooldown, and then just set the normal CD to that value when called?

Polynomix:
>Do we really need every Spell struct to hold a new CD timer for this to work? Wouldn't it be cleaner to just implement a generalized action like SetSpellCD(), which takes in both the spell and the time it should come off cooldown, and then just set the normal CD to that value when called?

So would you say that this should be useable on all spells or just the ones from gear ?

Polynomix:
I don't like the idea of allowing to do actions that are not available in the game. Is there any other use case where we'd want to set the CD to another value than the cooldown of the spell ?

NerdEgghead:
> I don't like the idea of allowing to do actions that are not available in the game. Is there any other use case where we'd want to set the CD to another value than the cooldown of the spell ?

This gave me a thought: can't you already accomplish this functionality by just casting the spell in a pre-pull action further in advance? So if you want to put Synapse Springs on a 30s CD, you would just cast it 30s before the pull in the APL, without the need for a custom action.

Polynomix:
This is assuming no trinkets effect will affect any prepull actions. I believe the gear swap interface right now is only for weapon swap, i'll look more into it after progress

PR #1194, opened 2024-11-11
[Balance] T13

stn-ms:
@rosenrusinov not sure you remember, but we spoke about this being implemented as DDBC a while ago. Here it is, let me know what you think.


sim/druid/items.go
```diff
@@ -197,14 +197,61 @@ var ItemSetDeepEarthRegalia = core.NewItemSet(core.ItemSet{
...
+
+			druid := agent.(DruidAgent).GetDruid()
+
+			t13InsectSwarmBonusDummyAuras := druid.NewEnemyAuraArray(func(target *core.Unit) *core.Aura {
+				return target.GetOrRegisterAura(core.Aura{
+					ActionID: core.ActionID{SpellID: 105722},
+					Label:    "Item - Druid T13 Balance 2P Bonus (Insect Swarm) - " + druid.Label,
+					Duration: core.NeverExpires,
+					OnGain: func(aura *core.Aura, sim *core.Simulation) {
+						core.EnableDamageDoneByCaster(DDBC_2pcT13, DDBC_Total, druid.AttackTables[aura.Unit.UnitIndex], t13InsectSwarmBonus)
```
NerdEgghead, line 222: Since this class only has a single `DynamicDamageDoneByCaster` that can ever be active, it will be cleaner and more performant to not mess with allocating and indexing a `DamageDoneByCasterExtraMultiplier` slice, and instead just set `DamageDoneByCasterMultiplier` directly:
```druid.AttackTables[aura.Unit.UnitIndex].DamageDoneByCasterMultiplier = t13InsectSwarmBonus```

File: sim/druid/items.go [hunk 2]
```diff
@@ -197,14 +197,61 @@ var ItemSetDeepEarthRegalia = core.NewItemSet(core.ItemSet{
...
+					},
+					OnExpire: func(aura *core.Aura, sim *core.Simulation) {
+						core.DisableDamageDoneByCaster(DDBC_2pcT13, druid.AttackTables[aura.Unit.UnitIndex])
```
NerdEgghead, line 225: Likewise, this can be simplified to:
```druid.AttackTables[aura.Unit.UnitIndex].DamageDoneByCasterMultiplier = nil```

File: sim/druid/items.go [hunk 3]
```diff
@@ -197,14 +197,61 @@ var ItemSetDeepEarthRegalia = core.NewItemSet(core.ItemSet{
 	Bonuses: map[int32]core.ApplyEffect{
 		// Insect Swarm increases all damage done by your Starfire, Starsurge, and Wrath spells against that target by 3%
 		2: func(agent core.Agent) {
+			const (
+				DDBC_2pcT13 int = iota
+
+				DDBC_Total
+			)
```
NerdEgghead, line 204: These constants can likewise be removed if you set `DamageDoneByCasterMultiplier` for each `AttackTable` directly. Furthermore, even in the event that additional `DynamicDamageDoneByCaster` modifiers need to be added for the class down the road, the indexing constants supplied to `EnableDamageDoneByCaster()` need to be defined at the top level of the `druid` package so that they can be accessed by all aura implementations that need them (similar to the class spell mask constants).

sim/druid/items.go
```diff
@@ -197,14 +197,61 @@ var ItemSetDeepEarthRegalia = core.NewItemSet(core.ItemSet{
...
+				})
+			})
+
+			druid.OnSpellRegistered(func(spell *core.Spell) {
+				if spell.ClassSpellMask&(DruidSpellInsectSwarm) == 0 {
+					return
+				}
+
+				for _, target := range druid.Env.AllUnits {
+					if target.Type == core.EnemyUnit {
```
NerdEgghead, line 236: This can be simplified to:
```for _, target := range druid.Env.Encounter.TargetUnits {```

NerdEgghead (changes requested):

NerdEgghead (approved):

PR #1209, opened 2024-11-15
Add separate APL Validation Warnings for each APL Value

raethkcj:
Adds the ability to attach validation warnings to individual APL Values that trigger them, rather than attaching them all to the parent Action.

Does this by assigning a UUID to each APLValue that matches its input's CSS ID, which gets sent to the backend, validated, and returned to the frontend with warnings if they exist.

Seemed like less traffic then putting the warnings themselves in APLValues and returning the whole APLRotation :shrug:

Also renames Warnings to Validations and adds a log-level-ish enum to them to label as Warning, Info, etc. and display them with a different icon+message on the frontend


1337LutZ:
**Consider renaming Warnings to Messages, and adding a log-level-ish enum to them to label as Warning, Info, etc. and display them with a different icon+message on the frontend**
Yes please!

ui/core/components/pickers/list_picker.tsx
```diff
@@ -383,6 +398,8 @@ export class ListPicker<ModObject, ItemType> extends Input<ModObject, Array<Item
 
 		itemHeaderElem.appendChild(validationElem);
 
+		const iconElem = validationElem.getElementsByTagName('i')[0];
```
1337LutZ, line 401: Please use `validationElem.querySelector('i')`

ui/scss/core/components/individual_sim_ui/_apl_rotation_picker.scss
```diff
@@ -57,6 +57,30 @@
 		width: unset;
 		margin: 0;
 	}
+
+	.apl-validation-information {
+		color: var(--bs-off-white) !important;
+		text-shadow:
+		0 0 var(--spacer-2) var(--bs-link-color),
+		0 0 var(--spacer-2) var(--bs-link-color),
+		0 0 var(--spacer-2) var(--bs-link-color);
```
1337LutZ, line 66: Why are there 3 shadows stacked on top of eachother? :O

raethkcj: :sweat_smile: I copied them from here without reading the docs https://github.com/wowsims/cata/blob/910e5ce1caee314690b24ae64547b2ae7aa9dcf2/ui/scss/shared/_global_old.scss#L91-L97

raethkcj: Changing to 1 shadow (left) looks better for Info but does make the shadow on Warning pretty faint compared to live (right):
image

raethkcj: I didn't end up using Error in any current validations, but here's what it looks like single versus triple (also open to feedback on color scheme lol)

image
image

Issue #1208, opened 2024-11-15
'Suggest Reforges' sometimes removes all reforges


stako:
**Describe the bug**
Clicking the 'suggest reforges' button sometimes removes all reforges

**To Reproduce**
1. Restore Defaults
2. Use WowSimsExporter addon to import gear/talents/etc
3. Click 'Suggest Reforges'
4. After some loading time, reforges are all removed

**Expected behavior**
After clicking 'Suggest Reforges' expected to have gear reforged approriately.

**Sim Links and Screenshots**
[...]

PR #1210, opened 2024-11-16
Suggest Reforges timeout fix for Firefox

NerdEgghead:
Changed timeout condition for Suggest Reforges LP solver from time-based (10s) to iteration-based (100k). This allows a partially converged solution to still be found even on low performance browsers (Firefox) - the solve will just take a lot longer compared to better browsers.

 On branch auto_reforge
 Changes to be committed:
	modified:   ui/core/components/suggest_reforges_action.tsx


NerdEgghead:
Fixes #1208

PR #1224, opened 2024-11-24
[WARLOCK] Add T13 set bonuses

1337LutZ:


sim/warlock/items.go
```diff
@@ -220,3 +221,48 @@ var ItemSetGladiatorsFelshroud = core.NewItemSet(core.ItemSet{
...
+		2: func(agent core.Agent) {
+			warlock := agent.(WarlockAgent).GetWarlock()
+
+			warlock.AddStaticMod(core.SpellModConfig{
+				Kind:      core.SpellMod_Cooldown_Flat,
+				TimeValue: -time.Minute * 4,
+				ClassMask: WarlockSpellSummonDoomguard | WarlockSpellSummonInfernal,
+			})
+
+			warlock.Tier132PSummonDuration = core.TernaryInt32(warlock.Spec == proto.Spec_SpecDemonologyWarlock, 20, 30)
```
Tharre, line 238: Nit but it's better calling the boni 2pT13 so that there's seperation between the digits

sim/warlock/soul_fire.go
```diff
@@ -56,6 +58,10 @@ func (warlock *Warlock) registerSoulFire() {
 			baseDamage := warlock.CalcAndRollDamageRange(sim, 2.54299998283, 0.22499999404)
 			result := spell.CalcDamage(sim, target, baseDamage, spell.OutcomeMagicHitAndCrit)
 
+			if hasT134P && warlock.SoulBurnAura.IsActive() {
+				warlock.SoulShards += 1
```
Tharre, line 62: Needs to be `min(3, warlock.Soulshards + 1)` since you can't have more then 3 shards at once.

sim/warlock/doomguard.go
```diff
@@ -9,7 +9,7 @@ import (
 )
 
 func (warlock *Warlock) registerSummonDoomguard(timer *core.Timer) {
-	duration := time.Duration(45+10*warlock.Talents.AncientGrimoire) * time.Second
+	duration := time.Duration(45+10*warlock.Talents.AncientGrimoire+warlock.Tier132PSummonDuration) * time.Second
```
Tharre, line 12: It would be nicer to make the tier bonus a helper function in items that calculates the summon duration extension instead of having a variable in the warlock struct that's only ever going to be written once.

Tharre (changes requested):

Tharre (approved):
LGTM

PR #1234, opened 2024-11-28
[Balance] Fixed a bug that allows Euphoria to proc outside of energy …

stn-ms:
…bounds at the start of an encounter


1337LutZ (approved):

1337LutZ:
Don't forget the tests 🙌

PR #1237, opened 2024-11-30
[Warrior] Implement charge as an actual movement out and back in

hillerstorm:
* Movement speed bonuses like enchants, meta gems and talents are now ExclusiveEffects
* Added MaxRange to all relevant Warrior abilities
* Changed Charge to require 8 yards and move back in rapidly when used
* Tweaked Arms default starting distance to 9 and changed pre-pull to use Charge a tiny bit earlier, to have it connect around 0
  * This change didn't affect the dps from before, which was the goal
* Added *hidden-by-default* APL actions for using Charge in the APL, for Charge-weave testing
  * Instead of doing this as a strict sequence, I separated the two and put the Charge action at the top to always use it if it detects that it's out of melee range.


1337LutZ:
LGTM but would like @NerdEgghead to check the code ;D

1337LutZ (approved):

sim/druid/feral/TestFeral.results
```diff
@@ -641,8 +641,8 @@ dps_results: {
 dps_results: {
  key: "TestFeral-AllItems-FleetShadowspiritDiamond"
  value: {
-  dps: 31809.54569
-  tps: 43875.63602
+  dps: 31851.34871
+  tps: 43729.52774
  }
 }
```
NerdEgghead, line 647: Feral Swiftness incorrectly stacking with the meta gem is likely why there is a test result change here. I think the right way to solve this is to register an `ExclusiveEffect` category for passive movement speed buffs in `core`, and then call the same `ExclusiveEffect` wrappers that we use for non-stacking buffs/debuffs when registering the enchants, meta gem, and Feral Swiftness effect, just with different priority levels so that the strongest one is applied.

sim/paladin/talents_retribution.go
```diff
@@ -46,6 +47,16 @@ func (paladin *Paladin) applyRuleOfLaw() {
 	})
 }
 
+func (paladin *Paladin) applyPursuitOfJustice() {
+	if paladin.Talents.PursuitOfJustice == 0 {
+		return
+	}
+
+	if paladin.PseudoStats.MovementSpeedMultiplier < 1.15 {
+		paladin.PseudoStats.MovementSpeedMultiplier = 1.15
```
NerdEgghead, line 56: This should also be done with `core.ExclusiveEffect` to properly handle the effects it doesn't stack with.

sim/warrior/charge.go
```diff
@@ -6,28 +6,82 @@ import (
 	"github.com/wowsims/cata/sim/core"
 )
 
+func performCharge(sim *core.Simulation, warrior *Warrior, metrics *core.ResourceMetrics, rage float64) {
+	warrior.AddRage(sim, rage, metrics)
+	if warrior.DistanceFromTarget > core.MaxMeleeRange {
+		warrior.PseudoStats.MovementSpeedMultiplier *= 2
```
NerdEgghead, line 12: I would assume that it's much more than a 2x speed increase? The way we measured this for Feral Charge is by logging dummy tests from variable target distances, where the tester mashed a melee range special after pressing Charge. We then looked at the delay between the Charge cast and the special cast in logs across the data set to fit a linear correlation versus distance.

sim/warrior/charge.go
```diff
@@ -6,28 +6,82 @@ import (
...
+
+	warrior.RegisterMovementCallback(func(sim *core.Simulation, position float64, kind core.MovementUpdateType) {
+		if warrior.ChargeAura.IsActive() && kind == core.MovementEnd {
+			if warrior.DistanceFromTarget >= chargeMinRange {
+				// Has moved out, charge back in
+				performCharge(sim, warrior, metrics, rage)
+			} else {
+				// Has charged back in, reset movement speed
+				warrior.PseudoStats.MovementSpeedMultiplier /= 2
+				warrior.ChargeAura.Deactivate(sim)
```
NerdEgghead, line 38: It's cleaner to put the movement speed changes within the `OnGain` and `OnExpire` callbacks for the Charge buff aura directly.

sim/warrior/charge.go
```diff
@@ -6,28 +6,82 @@ import (
...
+				// Always charge from min range in prepull
+				warrior.DistanceFromTarget = chargeMinRange
+			}
+
+			if warrior.DistanceFromTarget >= chargeMinRange {
+				performCharge(sim, warrior, metrics, rage)
+				return
+			}
 
+			warrior.MoveTo(chargeMinRange, sim)
```
NerdEgghead, line 84: Ideally, the `MoveTo()` should be done within the APL rather than as part of the Charge cast, or else you're conflating a rotational decision (moving out when preparing for a Charge) with what the spell is actually doing. You can look at the Feral rotation code for an example of this: even though the APL there is hard-coded, the moving out logic is done as part of the rotation code (i.e. the APL) rather than as part of the Charge cast, and the Feral Charge spell itself simply specifies the appropriate `MinRange` and `MaxRange` constraints for usage.

NerdEgghead (changes requested):

sim/warrior/charge.go
hillerstorm: I did some testing in game with logging turned on and it was very consistent with 3x so I changed it to that.
It's *almost* instant in the timeline, like 0.2-0.3s to charge back in.

hillerstorm:
@zoristaken please check if you have some time :)

NerdEgghead (approved):
Nicely done!

PR #1241, opened 2024-12-02
[UI] Feature/improve detailed results performance

1337LutZ:
- Added the request ID to the raid sim request to have a unique value to build cache keys with
- Added a generic CacheHandler to allow for caching of elements that do not change and can be reused.
- Added caching to diverse "enriched" HTML Nodes that are dependent on async calls like wowhead datasets
- Added caching (max 2 entries) to SimResults to make toggling between 2 (reference) datasets much quicker
- Added caching to the Timeline content
- Fixed DPS graph Timeline tooltips

**After & Before**
 ~ 3x performance increase when swapping/toggling between 2 results
Overall GC improvement
image

### DIFF
--- a/proto/api.proto
+++ b/proto/api.proto
@@ -360,6 +360,7 @@ message ErrorOutcome {
 
 // RPC RaidSim
 message RaidSimRequest {
+	string request_id = 5;
 	Raid raid = 1;
 	Encounter encounter = 2;
 	SimOptions sim_options = 3;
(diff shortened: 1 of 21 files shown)


hillerstorm (approved):

ToxicKevinFerm (approved):

hillerstorm:
This should be mergeable right @1337LutZ ?

1337LutZ:
> This should be mergeable right @1337LutZ ?

Yes! If others tested it and found now issues 👌

PR #1247, opened 2024-12-04
[WARLOCK] DS:FG Change CastTime mod to CastSpeed

1337LutZ:
- Changed incorrect CastTime spell mod to modify players cast speed


Tharre (approved):
LGTM

PR #1265, opened 2024-12-14
Add useful APL values for Energy and Focus

hillerstorm:
Added the following APL values to enable some more tricky APLs:
* Max Energy
* Max Focus
* Energy Regen Per Second
* Focus Regen Per Second
* Time to Target Energy
* Time to Target Focus

The energy regen values I added in my Monk branch to port the simc APL and thought they might be useful on live right away.

Also reordered some resource APL values and enabled filtering to make the menu less cluttered.


ToxicKevinFerm (approved):
Cool! Maybe rename time to max like estimated time to max or something?
Otherwise looks good!

NerdEgghead (commented):
Looks great! To complete the package, do you mind adding one additional helper value which is `TimeToTargetEnergy`? This can ofc be calculated using math expressions, but so can `EnergyTimeToMax`, and `TimeToTargetEnergy` is in practice the more useful of the two values for pooling calculations.

hillerstorm:
> Looks great! To complete the package, do you mind adding one additional helper value `TimeToTargetEnergy` (which takes a float input)? This can ofc be calculated using math expressions, but so can `EnergyTimeToMax`, and `TimeToTargetEnergy` is in practice the more useful of the two values for pooling calculations.

Oh, great idea 😅 will add!

PR #1273, opened 2024-12-21
[UI] Improve drag-and-drop support in APL editor

hillerstorm:
Doing a ton of APL editing highlighted a few missing features that would really speed things up, so I added them!

* Drag and drop APL values inside an `Any of` / `All of`
  * Not restricted to the same list, can be moved in/out any number of levels **and even between different APL actions**
* Being able to copy an APL value inside an `Any of` / `All of`
* Clearer indication of what block will be deleted inside nested `Any of` / `All of`
* Having the whole block show up as the drag-placeholder instead of just the move-button

Any feedback is appreciated!

https://github.com/user-attachments/assets/20f6edd3-4b19-4cfd-bee7-7a4b00f79ddf

### DIFF
--- a/ui/core/components/pickers/list_picker.tsx
+++ b/ui/core/components/pickers/list_picker.tsx
@@ -196,7 +196,13 @@ export class ListPicker<ModObject, ItemType> extends Input<ModObject, Array<Item
 
 		const itemHeader = document.createElement('div');
 		itemHeader.classList.add('list-picker-item-header');
-
+		
+		const popover = document.createElement('div');
+		popover.classList.add('list-picker-item-popover');
+		popover.setAttribute('popover', 'auto');
+		itemHeader.appendChild(popover);
+		let hasActions = false;
+		
 		if (this.config.inlineMenuBar) {
 			itemContainer.appendChild(itemElem);
 			itemContainer.appendChild(itemHeader);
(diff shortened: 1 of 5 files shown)


ui/core/components/pickers/list_picker.tsx
```diff
@@ -246,10 +282,14 @@ export class ListPicker<ModObject, ItemType> extends Input<ModObject, Array<Item
 			itemContainer.addEventListener(
 				'dragenter',
 				event => {
-					if (!curDragData || curDragData.listPicker != this) {
+					if (
+						!curDragData ||
+						curDragData.listPicker.config.itemLabel !== this.config.itemLabel ||
```
1337LutZ, line 287: Can this specific if check become a helper function? I see it repeated a lot (not that the old one wasn't)

ui/core/components/pickers/list_picker.tsx
```diff
@@ -274,30 +318,73 @@ export class ListPicker<ModObject, ItemType> extends Input<ModObject, Array<Item
...
+						(this.config.itemLabel === 'Action' && curDragData.listPicker !== this)
+					) {
+						return;
+					}
+					event.stopPropagation();
+					itemContainer.removeAttribute('draggable');
+					curDragData.item.elem.removeAttribute('draggable');
+					curDragData.item.elem.classList.remove('dragfrom');
+					[...document.querySelectorAll('.dragto,.hover')].forEach(elem => {
```
1337LutZ, line 349: Getting these specific values can probably also become a getter/util

ui/core/components/pickers/list_picker.tsx
hillerstorm: Yea I could do that 👍

ui/core/components/pickers/list_picker.tsx
hillerstorm: Do you mean the classes? Like a helper to fetch any number of elements matching certain classes and then remove the classes from the matching elements?
A very niche use case :D

hillerstorm: Cleaned up some repetitive code, this one being one of them

ui/core/components/pickers/list_picker.tsx
```diff
@@ -277,16 +281,18 @@ export class ListPicker<ModObject, ItemType> extends Input<ModObject, Array<Item
 				},
 				{ signal: this.signal },
 			);
+			
+			const invalidDropTarget = () => {
```
1337LutZ, line 285: Yes, this is what I meant :D

1337LutZ (approved):
I haven't tested the feature myself, but code wise it looks cleaner :D thanks for making some of the small util fns

Issue #1306, opened 2025-01-16
Crash Report 1174057138


Beardriuss:
Link:
[...]

Issue #1305, opened 2025-01-16
Crash Report 1644656809


morcano:
Link:
[...]

Issue #1303, opened 2025-01-16
Crash Report 742461799


HydraxSkarrag:
Link:
[...]

PR #1307, opened 2025-01-16
Add back prepull Mastery for secondary trinket swaps during swapped prepull

1337LutZ:
- Added prepull Mastery using Item Swap bonus stats


1337LutZ:
Fixes:
- #1308
- #1306
- #1305 
- #1303

PR #1313, opened 2025-01-19
Implement relative stat caps for "highest stat" trinket procs

NerdEgghead:


ui/core/proto_utils/gear.ts
```diff
@@ -163,6 +163,13 @@ export class Gear extends BaseGear {
 			.includes(itemId);
 	}
 
+	hasTrinketFromOptions(itemIds: number[]): boolean {
+		return this.getTrinkets()
+			.filter(t => !!t)
```
1337LutZ, line 168: A nicer way of doing this would be:
```
.filter((t): t is EquippedItem => !!t)
```
So you don't have to say `t!.item.id` but you just handle the type guard better

ui/core/components/suggest_reforges_action.tsx
```diff
@@ -859,6 +985,12 @@ export class ReforgeOptimizer {
 			this.setPseudoStatCoefficient(coefficients, PseudoStat.PseudoStatSpellHitPercent, appliedAmount);
 		}
 
+		// If a highest Stat constraint is to be enforced, then update the associated
+		// coefficient if applicable.
+		if (this.relativeStatCap) {
+			this.relativeStatCap!.updateCoefficients(coefficients, stat, amount);
```
1337LutZ, line 991: You could just do: `this.relativeStatCap?.updateCoefficients(coefficients, stat, amount);` so it will only execute when it's defined. Otherwise the `!` is not needed since you null check in the if.

ui/core/components/suggest_reforges_action.tsx
```diff
@@ -880,13 +1012,18 @@ export class ReforgeOptimizer {
...
+	buildYalpsConstraints(gear: Gear, baseStats: Stats): YalpsConstraints {
 		const constraints = new Map<string, Constraint>();
 
 		for (const slot of gear.getItemSlots()) {
 			constraints.set(ItemSlot[slot], lessEq(1));
 		}
 
+		if (this.relativeStatCap) {
+			const statsWithoutBaseMastery = baseStats.addStat(Stat.StatMasteryRating, -this.player.getBaseMastery() * Mechanics.MASTERY_RATING_PER_MASTERY_POINT);
+			this.relativeStatCap!.updateConstraints(constraints, gear, statsWithoutBaseMastery);
```
1337LutZ, line 1024: Same here, the `!` shouldn't be needed due to the null check in the if`

hillerstorm (approved):

PR #1331, opened 2025-01-30
Re-use rotation action instead of re-allocating

NerdEgghead:
When profiling melee sim runs, I observed that a considerable amount of overhead is spent on creating new `PendingAction` structs each time the rotation action is rescheduled (either after a cast, or after a wait call, etc.). This PR should make it safe to re-use cancelled `PendingAction` structs by removing them from the sim queue automatically during calls to `Cancel()`.

### DIFF
--- a/sim/core/gcd.go
+++ b/sim/core/gcd.go
@@ -57,18 +57,12 @@ func (unit *Unit) SetRotationTimer(sim *Simulation, rotationReadyAt time.Duratio
 
 	unit.RotationTimer.Set(rotationReadyAt)
 
-	if unit.rotationAction.consumed {
-		unit.rotationAction.cancelled = false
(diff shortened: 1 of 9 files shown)


NerdEgghead:
@TheBackstabi I'm pretty sure the Combat Rogue test changes are just an RNG effect due to the weird way it was set up before, I couldn't measure any true DPS difference relative to `master` on a 500k iteration sim with default settings. But tagging you for approval just in case I broke something unintentionally.

TheBackstabi (approved):
Yeah I'd assume just adjusting that GCD trigger might make things happen _slightly_ faster than they were before. It's not enough of a change that I'd be concerned. Killing Spree was unchanged going into Cata so I didn't adjust anything in its code, this seems better to me.

Also it made the numbers go up. Number go up good.

PR #1332, opened 2025-01-30
[MAGE] Optimize combustion APL performance

1337LutZ:
Following impact:
- 1.000.000 iterations on 18 cores from 01:16 to 00:53
- ~3 DPS loss due to slight desync in APL check vs Combust estimation update

### DIFF
--- a/sim/common/cata/other_effects.go
+++ b/sim/common/cata/other_effects.go
@@ -1415,6 +1415,7 @@ type IgniteDamageCalculator func(result *core.SpellResult) float64
 
 type IgniteConfig struct {
 	ActionID           core.ActionID
+	ClassSpellMask     int64
 	DisableCastMetrics bool
 	DotAuraLabel       string
 	DotAuraTag         string
@@ -1435,6 +1436,7 @@ func RegisterIgniteEffect(unit *core.Unit, config IgniteConfig) *core.Spell {
 		ActionID:         config.ActionID,
 		SpellSchool:      core.SpellSchoolFire,
 		ProcMask:         core.ProcMaskSpellProc,
+		ClassSpellMask:   config.ClassSpellMask,
 		Flags:            spellFlags,
 		DamageMultiplier: 1,
 		ThreatMultiplier: 1,
--- a/sim/core/unit.go
+++ b/sim/core/unit.go
@@ -11,7 +11,7 @@ import (
 type UnitType int
 type SpellRegisteredHandler func(spell *Spell)
 type OnMasteryStatChanged func(sim *Simulation, oldMasteryRating float64, newMasteryRating float64)
-type OnCastSpeedChanged func(oldSpeed float64, newSpeed float64)
+type OnCastSpeedChanged func(sim *Simulation, oldSpeed float64, newSpeed float64)
 type OnTemporaryStatsChange func(sim *Simulation, buffAura *Aura, statsChangeWithoutDeps stats.Stats)
 
 const (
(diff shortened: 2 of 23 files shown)


sim/common/cata/other_effects.go
```diff
@@ -1435,6 +1436,7 @@ func RegisterIgniteEffect(unit *core.Unit, config IgniteConfig) *core.Spell {
 		ActionID:         config.ActionID,
 		SpellSchool:      core.SpellSchoolFire,
 		ProcMask:         core.ProcMaskSpellProc,
+		ClassSpellMask:   config.ClassSpellMask,
```
1337LutZ, line 1439: I added this because sometimes (like in my case) you'd want to actually give the Ignite spell a mask

sim/core/unit.go
```diff
@@ -536,7 +536,7 @@ func (unit *Unit) finalize() {
 
 	unit.defaultTarget = unit.CurrentTarget
 	unit.applyParryHaste()
-	unit.updateCastSpeed()
+	unit.updateCastSpeed(nil)
```
1337LutZ, line 539: This will only have impact for the initial callback which (in this case) has no added value anyways

NerdEgghead: I don't like this because you're now requiring that every `OnCastSpeedChanged` implementation in spec packages has to do an explicit `nil` check on the `sim` input, which is not intuitive whatsoever. I don't think you actually need a `sim` input on the callback for your code to work at all, you're only using it for logging purposes. But if you do want the `sim` input (which is fine to match how the melee version works), then you should be able to simply delete this call, because all cast speed updates by construction will now be happening via auras that call `MultiplyCastSpeed()`, so no final processing should be required during `finalize()`.

So TLDR:
- If you want to force cast speed updates to always happen via Auras that call `MultiplyCastSpeed()` with a `sim` input, then you should just delete this final update call, and remove any explicit `nil` checks in callbacks.
- Alternatively, if you don't want to force this and want to retain flexibility for manually calling `MultiplyCastSpeed()` without a dynamic `Aura` in spec code, then you should keep the callback as a static one, and simply remove the sim-dependent logging in the Mage code.

sim/mage/combustion.go
```diff
@@ -113,4 +114,50 @@ func (mage *Mage) registerCombustionSpell() {
...
+			if sim.Log != nil {
+				mage.Log(sim, "Combustion Dot Estimate: %d - %d Ticks ", combustionDotDamage, combustionTickCount)
+			}
+		}
+
+		return combustionDotDamage
+	}
+
+	mage.AddOnCastSpeedChanged(func(sim *core.Simulation, old float64, new float64) {
+		if sim != nil {
```
NerdEgghead, line 141: Basically anytime you have to write a check like this, it means you should rethink how the code is being structured. Callbacks like this should never be called with `nil` inputs or you're just asking for developer headaches and weird bugs down the road.

sim/paladin/talents_retribution.go
```diff
@@ -73,13 +73,16 @@ func (paladin *Paladin) applySanctityOfBattle() {
...
+	paladin.AddOnCastSpeedChanged(func(_ *core.Simulation, _ float64, castSpeed float64) {
 		updateTimeValue(castSpeed)
 	})
 
 	core.MakePermanent(paladin.GetOrRegisterAura(core.Aura{
 		Label:    "Sanctity of Battle",
 		ActionID: core.ActionID{SpellID: 25956},
+		OnInit: func(aura *core.Aura, sim *core.Simulation) {
+
+		},
```
NerdEgghead, line 85: Is this needed?

NerdEgghead (changes requested):

sim/core/unit.go
1337LutZ: You're right. I've solved the logging in the APL Value itself, and leave this part only to do the calculation and updating of the value :)

NerdEgghead (approved):

PR #1360, opened 2025-02-08
Migrate Spell Costs to Cata

1337LutZ:
- Makes multipliers Int to prevent multithreading rounding errors
- Fixes a few bugged implementations
- Fix [Tyrande's Favorite Doll]
- Item Swap compatible [Shard of Woe]
- Add ResourceType to SpellMod to be able to deduct specific Power types


sim/core/spell.go
```diff
@@ -673,10 +682,27 @@ type SpellCost interface {
...
+	SpellCostFunctions
 }
 
-func (spell *Spell) EffectiveCritDamageMultiplier() float64 {
-	return (spell.CritMultiplier-1)*(spell.CritMultiplierAddative+1) + 1
+func (sc *SpellCost) ApplyCostModifiers(cost float64) float64 {
+	spell := sc.spell
+	cost = max(0, cost+float64(sc.FlatModifier))
+	cost = max(0, math.Floor(cost*float64(spell.Unit.PseudoStats.CostMultiplier)/100))
+	cost = max(0, math.Floor(cost*float64(sc.Multiplier)/100))
```
NerdEgghead, line 697: Do we know for sure that these get floored in game?

sim/core/spell_mod.go
```diff
@@ -65,18 +69,23 @@ func buildMod(unit *Unit, config SpellModConfig) *SpellMod {
 		removeFn = functions.Remove
 	}
 
+	if config.ResourceType > 0 && (config.ResourceType&proto.ResourceType_ResourceTypeMana|proto.ResourceType_ResourceTypeEnergy|proto.ResourceType_ResourceTypeRage|proto.ResourceType_ResourceTypeFocus == 0) {
```
NerdEgghead, line 72: Resource type enums are not bit masks, since they are incremented by 1 integer value rather than by 1 bit, so I don't think this check will ever fail.

sim/core/spell_mod.go
```diff
@@ -65,18 +69,23 @@ func buildMod(unit *Unit, config SpellModConfig) *SpellMod {
...
+		panic("ResourceType " + strconv.Itoa(int(config.ResourceType)) + " for SpellMod is not implemented")
+	}
```
NerdEgghead, line 74: I think a simple `%s` on `config.ResourceType` should work for a descriptive error message, since `ResourceType` implements `String()`.

sim/core/spell_mod.go
```diff
@@ -406,19 +427,27 @@ func removeDamageDoneAdd(mod *SpellMod, spell *Spell) {
...
 }
 
 func removePowerCostFlat(mod *SpellMod, spell *Spell) {
-	spell.DefaultCast.Cost -= mod.floatValue
+	if spell.Cost != nil {
+		spell.Cost.FlatModifier -= int32(mod.intValue)
+	}
 }
 
 func applyRunicPowerCostFlat(mod *SpellMod, spell *Spell) {
```
NerdEgghead, line 453: Can't this be removed now and wrapped into `applyPowerCostFlat()` with the appropriate `ResourceType`?

sim/druid/ferocious_bite.go
```diff
@@ -60,7 +60,7 @@ func (druid *Druid) registerFerociousBiteSpell() {
 			result := spell.CalcAndDealDamage(sim, target, baseDamage, spell.OutcomeMeleeSpecialHitAndCrit)
 
 			if result.Landed() {
-				druid.SpendEnergy(sim, excessEnergy, spell.Cost.(*core.EnergyCost).ResourceMetrics)
+				druid.SpendEnergy(sim, excessEnergy, spell.Cost.SpellCostFunctions.(*core.EnergyCost).ResourceMetrics)
 				druid.ApplyFeral4pT12(sim)
 				druid.SpendComboPoints(sim, spell.ComboPointMetrics())
```
NerdEgghead, line 65: Is there an analogous helper to `spell.ComboPointMetrics()` for `spell.EnergyMetrics()` that could be called here?

sim/shaman/unleash_elements.go
```diff
@@ -142,7 +142,7 @@ func (shaman *Shaman) registerUnleashLife() {
 		},
 		OnExpire: func(aura *core.Aura, sim *core.Simulation) {
 			for _, spell := range affectedSpells {
-				spell.CostMultiplier -= 0.2
+				spell.DamageMultiplierAdditive -= 0.2
```
NerdEgghead, line 145: Just double checking that this was intentional?

NerdEgghead (changes requested):
Would also recommend running a performance comparison before vs. after to make sure there isn't any overhead caused by this refactor.

sim/core/spell.go
1337LutZ: No, but I have just inherited the old code. SoD didn't have it, but Cata seemed to explicitely floor these values.

sim/core/spell_mod.go
1337LutZ: Ah yea I actually can, just need to look for the ResourceType 👍

sim/shaman/unleash_elements.go
1337LutZ: Yes, because the `OnGain` was never reverted in the `OnExpire`. I double checked the talent.

sim/core/spell_mod.go
1337LutZ: I tried, but it can't. 
RunicPower + Runes have a different way of tracking cost and currently this directly modifies the float value of the cost. It never actually uses additive/multiplicative modifers. 
As my intent is not to fully refactor the way DK RP/Runes are spent I rather leave it as is

sim/core/spell_mod.go
```diff
@@ -199,12 +224,12 @@ const (
 	// Uses FloatValue
 	SpellMod_DamageDone_Flat
 
-	// Will reduce spell.DefaultCast.Cost by % amount. -5% = -0.05
-	// Uses FloatValue
+	// Will reduce spell.Cost.Multiplier by % amount. -5% = -5
+	// Uses IntValue
 	SpellMod_PowerCost_Pct
```
NerdEgghead, line 229: Should be `spell.Cost.PercentModifier` instead of `spell.Cost.Multiplier`

NerdEgghead (changes requested):

sim/core/spell_mod.go
```diff
@@ -406,19 +431,27 @@ func removeDamageDoneAdd(mod *SpellMod, spell *Spell) {
 }
 
 func applyPowerCostPercent(mod *SpellMod, spell *Spell) {
-	spell.CostMultiplier += mod.floatValue
+	if spell.Cost != nil {
+		spell.Cost.PercentModifier += int32(mod.intValue)
+	}
```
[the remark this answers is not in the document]
1337LutZ: No, I don't think we have any place where `intValue` would ever be close to exceeding the `int32` value.

sim/core/spell_mod.go
```diff
@@ -106,6 +117,18 @@ func shouldApply(spell *Spell, mod *SpellMod) bool {
 		return false
 	}
 
+	if mod.ResourceType > 0 && spell.Cost != nil {
```
NerdEgghead, line 120: Probably cleaner to also return `false` if `(mod.ResourceType > 0) && (spell.Cost == nil)`.

NerdEgghead (changes requested):
Almost there, just a few pending issues + resolving merge conflicts for a clean unit test comparison vs. `master`.

sim/core/mana.go
```diff
@@ -297,33 +297,27 @@ func (mb *manaBar) EndOOMEvent(sim *Simulation) {
...
-	}
-
-	spell.DefaultCast.Cost = math.Floor(baseCost * TernaryFloat64(options.Multiplier == 0, 1, options.Multiplier))
-
-	return &ManaCost{
-		ResourceMetrics: spell.Unit.NewManaMetrics(spell.ActionID),
+func newManaCost(spell *Spell, options ManaCostOptions) *SpellCost {
+	return &SpellCost{
+		spell:           spell,
+		BaseCost:        TernaryInt32(options.FlatCost > 0, options.FlatCost, int32(math.Floor(float64(options.BaseCostPercent)/100*spell.Unit.BaseMana))),
```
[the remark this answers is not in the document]
1337LutZ: Ah I forgot about this one!

NerdEgghead (approved):

PR #1380, opened 2025-02-13
Gurthalak changes from PTR

hillerstorm:


sim/common/cata/gurthalak.go
```diff
@@ -62,42 +53,36 @@ func init() {
...
-				},
-			})
-
-			character.ItemSwap.RegisterProc(gurthalakItemID, aura)
+			if character.Equipment.MainHand().ID == gurthalakItemID {
+				makeGurthalakProcTrigger(character, gurthalakItemID, summonSpell, labelSuffix, true)
+			}
+			if character.Equipment.OffHand().ID == gurthalakItemID {
+				makeGurthalakProcTrigger(character, gurthalakItemID, summonSpell, labelSuffix, false)
+			}
```
NerdEgghead, line 61: This won't behave properly if Gurthalak is present in the swap set but not the main one, the Aura won't be registered at all in that scenario. I think you can fix it by doing
```
if (character.Equipment.MainHand().ID == gurthalakItemID) || (character.ItemSwap.GetUnequippedItemBySlot(proto.ItemSlot_ItemSlotMainHand).ID == gurthalakItemID) {
```
and likewise for the OH registration.

sim/common/cata/gurthalak.go
```diff
@@ -196,57 +179,45 @@ func (pet *TentacleOfTheOldOnesPet) registerMindFlay() {
...
-				}
 
 				if result.Landed() {
+					dot := spell.Dot(target)
 					dot.Apply(sim)
-					// In old videos there seems to be a delay between each cast, we simulate this by setting an explicit wait time.
-					// TODO: Verify this on the PTR
-					pet.castDelay = sim.CurrentTime + time.Second*4
+					// Delaying two NPC GCDs minus the total cast time of Mind Flay (3s)
+					pet.ExtendGCDUntil(sim, sim.CurrentTime+core.DurationFromSeconds(3.24))
```
NerdEgghead, line 188: Clearer to understand if you write it as `sim.CurrentTime + core.BossGCD*2`

sim/core/pet.go
```diff
@@ -220,6 +220,10 @@ func (pet *Pet) EnableWithTimeout(sim *Simulation, petAgent PetAgent, petDuratio
 	sim.AddPendingAction(pet.timeoutAction)
 }
 
+func (pet *Pet) SetStartattackDelay(startAttackDelay time.Duration) {
+	pet.startAttackDelay = startAttackDelay
+}
```
NerdEgghead, line 225: Minor nit, call it `SetStartAttackDelay()` for consistency with the variable name.

sim/common/cata/gurthalak.go
1337LutZ: I would probably add a function like we have for `CouldHaveSetBonus` called: `CouldHaveItem`

1337LutZ (approved):

Issue #1463, opened 2025-03-28
Add ability to disable stats for stat weight calculation


kayla-glick:
https://github.com/wowsims/sod/pull/1369

PR #1479, opened 2025-04-22
Implement #1463: ability to disable stat weights from calculation

1337LutZ:


ui/core/components/stat_weights_action.tsx
```diff
@@ -4,21 +4,110 @@ import { ref } from 'tsx-vanilla';
...
+			if (excludedStats?.length) this.excludedStats = excludedStats;
+			if (excludedPseudoStats?.length) this.excludedPseudoStats = excludedPseudoStats;
+			this.changeEmitter.emit(eventID);
+		}
+	}
+
+	toProto(): SavedStatWeightSettings {
+		const proto = SavedStatWeightSettings.create();
+		proto.excludedStats = this.excludedStats;
+		proto.excludedPseudoStats = this.excludedPseudoStats;
```
NerdEgghead, line 55: Slice these ones as well to make safe copies. Additionally, you should set `proto.apiVersion` here.

NerdEgghead (changes requested):
Much cleaner overall!

PR #1485, opened 2025-05-01
[Fix] Round spell mod after iteration

InDebt:
This should help mitigate errors inf loating point computation that might accumulate over iterations.
When a serial simulation of N iterations is performed spell mods are going to add add subtract to the respective field for every iteration. That field is never reset between iterations but relies on the assumption that `1 + 1 - 1 = 1`.

This however is not true for all combinations of floating point numbers and orders of execution. Small errors might leave the field in a state where it is now 0.9999999998 after an iteration. This error accumulates over thousands of iterations.

When a parallalel simulation is run, only a fraction of iterations is performed per instance of the simulation and the error will not manifest it self as pronounced, causing slight differences in the amounts calculated for higher iterations.

For now we noticed this issue only for addative damage mods but we can easily add more here if we wish for it.

### DIFF
--- a/sim/core/spell_mod.go
+++ b/sim/core/spell_mod.go
@@ -2,6 +2,7 @@ package core
 
 import (
 	"fmt"
+	"math"
 	"slices"
 	"strconv"
 	"time"
@@ -97,6 +98,16 @@ func buildMod(unit *Unit, config SpellModConfig) *SpellMod {
 			if mod.IsActive {
 				mod.Apply(mod, spell)
 			}
+		}
+	})
+
+	unit.RegisterResetEffect(func(s *Simulation) {
+		for _, spell := range unit.Spellbook {
+			spell.DamageMultiplierAdditive = math.Round(spell.DamageMultiplierAdditive*10000) / 10000
+
+			// Possibly add other spell mod variables here to safely round them
+			// Spell values are not reset on Iteration Reset so small floating point errors by multiplying / adding to a field
+			// Will carry over through iterations
 		}
 	})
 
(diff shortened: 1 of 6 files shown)


InDebt:
Updated implmentation to not register a reste function for each and every mod but move it into the actual mod to implement any reset logic between iterations. Now also correctly supports reset logic for custom mods.

sim/core/spell_mod.go
```diff
@@ -25,6 +26,7 @@ type SpellModConfig struct {
 	KeyValue     string
 	ApplyCustom  SpellModApply
 	RemoveCustom SpellModRemove
+	OnReset      SpellModOnReset
```
NerdEgghead, line 29: `ResetCustom` would be more consistent naming within `SpellModConfig`, since this field is ignored except when `config.Kind == SpellMod_Custom`.

sim/core/spell_mod.go
```diff
@@ -66,9 +73,16 @@ func buildMod(unit *Unit, config SpellModConfig) *SpellMod {
 
 		applyFn = config.ApplyCustom
 		removeFn = config.RemoveCustom
+
+		if config.OnReset != nil {
+			resetFn = config.OnReset
+		} else {
+			resetFn = functions.OnReset
+		}
```
NerdEgghead, line 81: This can be simplified to just
```
resetFn = config.ResetCustom
```
since `functions.OnReset` will just be `nil` in this case regardless.

NerdEgghead (commented):
This solution looks solid, just have a few minor suggestions.

PR #1491, opened 2025-05-08
Test/itemeffects

ToxicKevinFerm:


tools/database/dbc/item_effect.go
```diff
@@ -33,3 +40,193 @@ func (e *ItemEffect) ToMap() map[string]interface{} {
 func GetItemEffect(effectId int) ItemEffect {
 	return dbcInstance.ItemEffects[effectId]
 }
+
+var emptyStats = stats.Stats{}
```
NerdEgghead, line 44: Avoid using package-level global variables like this.

tools/database/dbc/item_effect.go
```diff
@@ -33,3 +40,193 @@ func (e *ItemEffect) ToMap() map[string]interface{} {
...
+	pe := newProtoShell(e)
+
+	statsSpellID, ilvlRppmMod := applyTrigger(e, pe, itemLevel)
+
+	pe.ScalingOptions[int32(levelState)] = buildScalingProps(statsSpellID, itemLevel)
+	if ilvlRppmMod != 0.0 && ilvlRppmMod != 1.0 {
+		pe.ScalingOptions[int32(levelState)].RppmIlvlModifier = ilvlRppmMod
+	}
+	return pe
+}
```
NerdEgghead, line 56: Don't you need to mirror the logic used for items, where you determine all of the relevant `ItemLevelState` settings that the effect could have and then populate `ScalingOptions` entries for each of those keys?

sim/core/flags.go
```diff
@@ -274,6 +274,11 @@ func SpellSchoolFromProto(p proto.SpellSchool) SpellSchool {
 	}
 }
 
+const (
+	RPPM_HASTE = 1 << (iota)
+	RPPM_CRIT
+)
```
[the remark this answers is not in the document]
NerdEgghead: Also do these definitions need to live in `core`, or can they exist solely within `dbc`?

tools/database/dbc/item_effect.go
```diff
@@ -33,3 +40,193 @@ func (e *ItemEffect) ToMap() map[string]interface{} {
...
+
+func buildScalingProps(spellID, itemLevel int) *proto.ScalingItemEffectProperties {
+	total := collectStats(spellID, itemLevel)
+	src := total.ToProtoMap()
+
+	m := make(map[int32]float64, len(src))
+	maps.Copy(m, src)
+
+	return &proto.ScalingItemEffectProperties{Stats: m}
```
NerdEgghead, line 166: Can you just directly populate the `Stats` field with `src` rather than allocating another map and copying to it first?

tools/database/dbc/item_effect.go
```diff
@@ -33,3 +40,193 @@ func (e *ItemEffect) ToMap() map[string]interface{} {
...
+		if base == nil {
+			continue
+		}
+
+		if len(base.ScalingOptions) > 0 {
+			merged = append(merged, base)
+		}
+	}
+	return merged
+}
```
NerdEgghead, line 232: Ok I see where you're handling the full set of `ItemLevelState` options now. This seems very inefficient though, since you're allocating and populating a whole other `proto.ItemEffect` struct for every variant of the item. Wouldn't it be better to create the base `proto.ItemEffect` only once per item, and then only update the `ScalingOptions` map within the loop over `ItemLevelState` options rather than merging a full proto?

tools/database/dbc/spell_effect.go
```diff
@@ -207,15 +207,34 @@ func (data *SpellEffect) ClassFlag(index uint) uint32 {
...
+func (effect *SpellEffect) ParseStatEffect(scalesWithIlvl bool, ilvl int) *stats.Stats {
 	stats := &stats.Stats{}
 	scale := effect.ScalingClass()
-	spell := dbcInstance.Spells[effect.ID]
+	spell := dbcInstance.Spells[effect.SpellID]
 	stat, _ := MapMainStatToStat(effect.EffectMiscValues[0])
 
 	switch {
+	case effect.EffectAura == A_MOD_RANGED_ATTACK_POWER:
+		if scalesWithIlvl && effect.Coefficient != 0 && spell.Attributes[8]&0x1000 != 0 {
```
NerdEgghead, line 218: Pre-cache this check before the switch rather than copy-pasting it three times:
```
useRandPropPoints := scalesWithIlvl && (effect.Coefficient != 0) && (spell.Attributes[8]&0x1000 != 0)
```

tools/database/dbc/spell_effect.go
```diff
@@ -207,15 +207,34 @@ func (data *SpellEffect) ClassFlag(index uint) uint32 {
...
+			propPoints := dbcInstance.RandomPropertiesByIlvl[ilvl][proto.ItemQuality_ItemQualityEpic][0] // Epic 0 for items
+			stats[proto.Stat_StatRangedAttackPower] = math.Floor(float64(propPoints) * effect.Coefficient)
```
NerdEgghead, line 220: Likewise, define a helper for this calculation so you're only coding the logic once:
```
func (effect *SpellEffect) CalcRandomPropStatValue(ilvl int) float64 {
    propPoints := dbcInstance.RandomPropertiesByIlvl[ilvl][proto.ItemQuality_ItemQualityEpic][0] // Epic 0 for items
    return math.Floor(float64(propPoints) * effect.Coefficient)
}
```

tools/database/dbc/spell_effect.go
```diff
@@ -247,7 +266,13 @@ func (effect *SpellEffect) ParseStatEffect() *stats.Stats {
 
 	case effect.EffectAura == A_MOD_RATING:
 		for _, rating := range getMatchingRatingMods(effect.EffectMiscValues[0]) {
+
 			if statMod := RatingModToStat[rating]; statMod != -1 {
+				if scalesWithIlvl && effect.Coefficient != 0 && spell.Attributes[8]&0x1000 == 0 {
```
NerdEgghead, line 271: You're doing `spell.Attributes[8]&0x1000 == 0` here when it's been `!= 0` the previous three times, is that intended?

NerdEgghead (commented):
Great start, this will hugely simplify our very messy on-use item factories once it's finished!

tools/database/dbc/item_effect.go
ToxicKevinFerm: Yeah I built myself into a bit of a corner when I realized I had to recursively step through spell triggers and also scale the stats array to the item level 😄

proto/db.proto
```diff
@@ -36,6 +36,7 @@ message SimItem {
 	int32 set_id = 12;
 
 	map<int32, ScalingItemProperties> scaling_options = 13; // keys are the all ItemLevelState variants that this item could potentially have
+	repeated ItemEffect item_effects = 14;
```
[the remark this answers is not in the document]
NerdEgghead: > @NerdEgghead I think trinkets like this:
> 
> * https://www.wowhead.com/mop-classic/item=105472/assurance-of-consequence
> * https://www.wowhead.com/mop-classic/item=105527/haromms-talisman
> * https://www.wowhead.com/mop-classic/item=105459/fusion-fire-core
> * and more

Trinkets like these will require custom implementations for the second effect regardless, it's only the simple stat effect that we can handle automatically with factory functions. If those are the only examples, then it would be better to just remove the complex effect during pre-processing and dump only the stat buff effect.

tools/database/dbc/item_effect.go
```diff
@@ -33,3 +39,199 @@ func (e *ItemEffect) ToMap() map[string]interface{} {
...
+		pe.Type = proto.ItemEffectType_EffectTypeOnUse
+		pe.Effect = &proto.ItemEffect_OnUse{OnUse: &proto.OnUseEffect{
+			Cooldown:         int32(e.CoolDownMSec / 1000),
+			CategoryId:       int32(e.SpellCategoryID),
+			CategoryCooldown: int32(e.CategoryCoolDownMSec / 1000),
+		}}
+	case ITEM_SPELLTRIGGER_CHANCE_ON_HIT:
+		proc := &proto.ProcEffect{
+			ProcChance: float64(spTop.ProcChance) / 100,
+			Icd:        int32(spTop.ProcCategoryRecovery / 1000),
```
InDebt, line 73: Is there a reason we do not use milliseconds directly here? Many spells have 500ms or 800ms or shorter ICD windows especially stacking auras. I'm only versed in older trinkets like https://www.wowhead.com/mop-classic/item=40430/majestic-dragon-figurine or https://www.wowhead.com/mop-classic/item=28785/the-lightning-capacitor but we'd not be able to easily port this code to SoD, TBC classic and so on

tools/database/dbc/item_effect.go
```diff
@@ -33,3 +39,199 @@ func (e *ItemEffect) ToMap() map[string]interface{} {
...
+func assignTrigger(e *ItemEffect, statsSpellID int, pe *proto.ItemEffect) {
+	spTop := dbcInstance.Spells[e.SpellID]
+	statsSP := dbcInstance.Spells[statsSpellID]
+	switch resolveTriggerType(e.TriggerType, e.SpellID) {
+	case ITEM_SPELLTRIGGER_ON_USE:
+		pe.Type = proto.ItemEffectType_EffectTypeOnUse
+		pe.Effect = &proto.ItemEffect_OnUse{OnUse: &proto.OnUseEffect{
+			Cooldown:         int32(e.CoolDownMSec / 1000),
+			CategoryId:       int32(e.SpellCategoryID),
+			CategoryCooldown: int32(e.CategoryCoolDownMSec / 1000),
```
InDebt, line 68: Same, I think we should store Milli Second values as milli seonds. There are simple on use effects like this https://wago.tools/db2/ItemEffect?build=5.5.0.60802&filter%5BParentItemID%5D=104298&page=1&sort%5BParentItemID%5D=desc that have a 1.5 sec (GCD) shared cooldown with other offensive spells

NerdEgghead: Likewise, we should rename these proto fields to `CooldownMs` and `CategoryCooldownMs` then so that the units are explicit.

tools/database/dbc/item_effect.go
```diff
@@ -33,3 +41,210 @@ func (e *ItemEffect) ToMap() map[string]interface{} {
...
+		if proc := pe.GetProc(); proc != nil {
+			_, specMods := realPpmModifier(dbcInstance.Spells[statsSpellID], ilvl)
+			proc.SpecModifiers = specMods
+		}
+	}
+
+	return pe
+}
+
+func realPpmScale(spell Spell) int {
```
NerdEgghead, line 219: You probably want to take in a `*Spell` pointer here instead, in order to avoid copying the input on every call.

tools/database/dbc/spell_effect.go
```diff
@@ -4,12 +4,13 @@ import (
...
 	"slices"
 
+	"github.com/wowsims/cata/sim/core"
 	"github.com/wowsims/cata/sim/core/proto"
 	"github.com/wowsims/cata/sim/core/stats"
 )
 
 const MAX_SCALING_LEVEL = 100
-const BASE_LEVEL = 85
+const BASE_LEVEL = 90
```
NerdEgghead, line 13: Is this maybe why you were getting the wrong stat values for the boosted trinkets?

tools/database/dbc/dbc.go
```diff
@@ -23,8 +23,9 @@ type DBC struct {
 	ItemArmorTotal         map[int]ItemArmorTotal
 	ArmorLocation          map[int]ArmorLocation
 	SpellScalings          map[int]SpellScaling
-	Consumables            map[int]Consumable // Item ID
-	ItemEffects            map[int]ItemEffect // Parent Item ID
+	Consumables            map[int]Consumable   // Item ID
+	ItemEffects            map[int]ItemEffect   // Parent Item ID
+	ItemEffectsByParentID  map[int][]ItemEffect // new
```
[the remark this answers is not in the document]
ToxicKevinFerm: No ItemEffects is keyed by effect id, the comment is wrong.
So, an item can have several effects and sometimes I need to be able to look up both. For example for consumables or or enchants.

tools/database/dbc/spell_effect.go
```diff
@@ -206,21 +200,45 @@ func (effect *SpellEffect) IsPeriodicDamageEffect() bool {
...
+func (effect *SpellEffect) GetScalingValue(ilvl int) float64 {
+	if ilvl > 0 {
+		// If item we get rand prop points
+		return float64(dbcInstance.RandomPropertiesByIlvl[ilvl][proto.ItemQuality_ItemQualityEpic][0])
+	}
+	spell := dbcInstance.Spells[effect.SpellID]
+	// if not we get class scaling based on the spell
 	scale := effect.ScalingClass()
-	spell := dbcInstance.Spells[effect.ID]
+	return dbcInstance.SpellScalings[core.TernaryInt(spell.MaxScalingLevel > BASE_LEVEL, BASE_LEVEL, spell.MaxScalingLevel)].Values[scale]
```
NerdEgghead, line 215: `core.TernaryInt(spell.MaxScalingLevel > BASE_LEVEL, BASE_LEVEL, spell.MaxScalingLevel)` can be simplified to just `min(spell.MaxScalingLevel, BASE_LEVEL)`.

NerdEgghead (changes requested):

sim/druid/guardian/TestGuardian.results
```diff
@@ -1144,15 +1144,15 @@ dps_results: {
...
+  hps: 70.70433
  }
 }
 dps_results: {
  key: "TestGuardian-AllItems-HeartofThunder-56370"
  value: {
   dps: 10601.44821
   tps: 53073.64564
-  hps: 355.12858
+  hps: 70.70433
```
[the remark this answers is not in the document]
ToxicKevinFerm: I assume its either due to the conjured cd change or the defensive cd change. I couldnt actually find the cause

NerdEgghead (approved):

PR #1515, opened 2025-06-26
Dynamic Encounter Targets

NerdEgghead:


proto/common.proto
```diff
@@ -835,6 +835,9 @@ message Target {
 	// Used in tank swap AIs.
 	int32 second_tank_index = 100;
 
+        // Used in dynamic target AIs.
```
dodoels, line 838: indent is off

sim/core/target.go
```diff
@@ -174,18 +214,83 @@ func (target *Target) Reset(sim *Simulation) {
...
 		nextIndex = 0
 	}
-	return target.Env.GetTarget(nextIndex)
+
+	nextTarget := target.Env.GetTargetByIndex(nextIndex)
+
+	if nextTarget.IsEnabled() {
+		return nextTarget
+	} else {
+		return nextTarget.NextActiveTarget()
```
ToxicKevinFerm, line 292: Won't this cause problems if every target is disabled for some reason? Shouldn't happen of course but never know.

ToxicKevinFerm (commented):
Looks good!

sim/core/target.go
NerdEgghead: This situation shouldn't be possible because of the `panic` I put in `removeInactiveTarget()`.

